PR #39, opened 2023-07-12
added an alternative to md2pdf(weasyprint)

gregdrizz:
added an alternative to md2pdf, since weasy print causes a lot of trouble in different operating systems(especially windows)

pdf kit uses wkhtmltopdf as an engine which is super to install and configure on any system, from UNIX based systems to windows.

Tested and seems to be working, windows users might have to add these lines if they don't add wkhtmltopdf/bin to their PATH:

path_wkhtmltopdf = r'C:\{INSTALLATION_DIRECTORY}\wkhtmltopdf\bin\wkhtmltopdf.exe'
config = pdfkit.configuration(wkhtmltopdf=path_wkhtmltopdf)
(won't be needed in most cases)


assafelovic:
Thank you @gregdrizz this is super helpful for users having issues with WeasyPrint. We're currently considering the best way forward and will take this in mind. Feel free to join our discussion on Discord in the meantime!

andrewDoing:
@assafelovic Any decision made on discord for this PR that should be posted here as well?

Issue #30, opened 2023-07-12
Remote (non localhost) address support


shon:
If one hosts it on public IP / address, then it throws the error `WebSocket connection to 'ws://localhost:8000/ws' failed:`
Changing `client/scripts.js` localhost value to my host's IP worked but obviously, it won't work for next update.
I started the app `uvicorn main:app --reload --host 0.0.0.0`
Can we support remote addresses please?

jortegac:
Can you try with the following?

`uvicorn main:app --host 0.0.0.0 --port 8000`

rotemweiss57:
Hi @shon it is in my backlog. feel free to work on it and add a pull request!

shon:
> Hi @shon it is in my backlog. feel free to work on it and add a pull request!

Done. Have a look https://github.com/assafelovic/gpt-researcher/pull/63
And thanks for this amazing tool

PR #63, opened 2023-07-15
Support non localhost setups

shon:
Replaced hardcoded localhost with dynamic ws url generation
Fixes: https://github.com/assafelovic/gpt-researcher/issues/30

### DIFF
--- a/client/scripts.js
+++ b/client/scripts.js
@@ -10,8 +10,10 @@ const startResearch = () => {
 }
 
 const listenToSockEvents = () => {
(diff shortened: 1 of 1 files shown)


assafelovic (approved):
Thank you this is great!

sam1am:
It looks like the request is succesful, but I never get a response when using a custom domain. 

Here I made two requests from my custom domain and one from localhost: 

Screenshot 2023-07-31 at 4 52 01 PM

It looks like the first two requests are entirely missing a host.

PR #72, opened 2023-07-16
Host on linux

ElishaKay:
The root Dockerfile is missing the Chrome & Chrome-Driver dependencies - so it will only run in dev mode (local machine). 

The fix in this branch will enable you to publish the Docker image on ubuntu. Without the fix, the scraper fails in Production.
From my research, Chrome & Chrome Driver should be the same version. (This branch contains the latest stable versions of both)

Tested locally on Windows - with VScode ssh-ed into server.
Tested on Ubuntu 22.10 - see image below for Digital Ocean droplet size.

You can view the demo with this branch at: https://gptresearcher.gptbud.com

digital-ocean-droplet


assafelovic (approved):
This is brilliant, thank you!

PR #84, opened 2023-07-19
name_debug

Ryoo72:
The image was not loading properly due to incorrect case, I fixed it.


assafelovic:
We've passed on this since the client has changed and does not present avatars anymore. Thank you!

PR #106, opened 2023-07-27
Add custom api base

paradoxtown:
I believe that including an option for a (free) third-party API (e.g., [chimeragpt](https://chimeragpt.adventblocks.cc/)) could be beneficial for those who have trouble paying for the openai api or just as an alternative for development.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,7 @@ $ pip install -r requirements.txt
 
 ```bash
 $ export OPENAI_API_KEY={Your API Key here}
+$ export OPENAI_API_BASE={Your API Base here} (required if you are using a custom API base)
 ```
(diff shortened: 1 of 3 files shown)


dreadful-dev:
If the custom option already has an openai compliant api, [this is possible](https://github.com/openai/openai-python/blob/5d50e9e3b39540af782ca24e65c290343d86e1a9/openai/__init__.py#L50) through the openai dependency. All you need to do is pop the following environment variable into a `.env` file or export it via bash

```
OPENAI_API_BASE=<OPENAI_COMPLIANT_API>
```

assafelovic:
Added Langchain adapters for more LLM options

PR #108, opened 2023-07-28
use langchain

hwchase17:
This uses langchain, values being:

1 - automatically works with langsmith out of the box

Screenshot 2023-07-28 at 12 16 57 PM

2 - sets it up to easily (relatively easily) start using other language models (need to discuss how exactly that should be configured)

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,10 @@ $ pip install -r requirements.txt
 ```bash
 $ export OPENAI_API_KEY={Your API Key here}
 ```
+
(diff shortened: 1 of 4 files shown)


hwchase17:
this needs to wait for the next version of langchain actually

gregdrizz (approved):

gregdrizz (changes requested):
@hwchase17 Thanks! 
We'll wait for the update and then merge :)

gregdrizz (approved):
Works with the new version of langchain(0.0.247)

hwchase17:
@gregdrizz yup - this should be good to go in whenever!

PR #112, opened 2023-07-31
Ability to select GPT models

zebleck:
Since its a bit annoying to have to change to the GPT-3.5-model in code, it might be useful to change the gpt model between gpt-3.5-turbo and gpt-4 in the frontend. Might also be nice if someone wants to save some credits and thus wants to switch from 4 to 3.5.

* Uses radio buttons for selection.
* Shows an error on failure in the agent output box.
* I also changed the Config to use the Singleton pattern (one instance for the whole application), which a Configuration should be.


assafelovic:
@zebleck super appreciate the contribution! However, we optimize various models of gpt across the algorithm therefore we would rather not have users select the model. Only in cases where users don't have access for GPT4 it would make sense. Thanks again, and feel free to check out our roadmap for additional contributions!

PR #150, opened 2023-08-23
Integrate w/ LLaMA

rlancemartin:
We use Ollama for [integration](https://python.langchain.com/docs/integrations/llms/ollama) w/ LLaMA:

* Follow setup instructions note in the linked docs.
 
Note: output formatting for lists requires particular prompts. 

* Good topic for prompt management @efriis 
  
I can get the app to run locally end-to-end w/ `LLaMA2-7b` as configured in this PR:
image

But, some notes as I hit a bunch of system errors running this:

* Many additional pkgs needed beyond pip installs:`brew install fontconfig, harfbuzz, pango, glib`
* Mismatch btwn `chrome` and `chromedriver` versions most likely throwing below error (seen in final [prompt](https://smith.langchain.com/public/2ce3af84-663d-4095-bfdf-55d362fc4a47/r)).
```
Message: Can not connect to the Service /Users/rlm/.cache/selenium/chromedriver/mac-arm64/116.0.5845.96/chromedriver
```
* But these are issues independent of LLaMA, and I'm not sure they are reproduced for others.
* In any case, the changes reflected here should enable running gpt-researcher locally w/ LLaMA.


jlcases:
Hi, i revised your code, but with this code, We cannot use it with OpenAI.... it makes more sense a selector, IMHO

PR #154, opened 2023-08-27
Fix copy to clipboard not working

Di3mex:
A missing `GPTResearcher.` before the function call prevented the copy to clipboard function from being called.


assafelovic (approved):
Thank you for the fix!

PR #173, opened 2023-09-09
Added module to autoupgrade and select the proper Chromedriver version

### LINKED ISSUE #171: Getting unauthorized software warning from apple on chromedriver
“chromedriver” cannot be opened because the developer cannot be verified.

Might be unique to my system, but we should add something to the readme perhaps?  I'll look into it

jacobSingh:
Closes #171 since the correct version is always installed and selected if unavailable.


assafelovic:
Very valuable contribution, thank you!

assafelovic (approved):

PR #181, opened 2023-09-10
Fixes the UX issue of download buttons showing before the report is done

### LINKED ISSUE #180: Download links should not be active until research is done
This kinda confused me and would probably confuse others.

Moving to react or vue and doing this with state management in the DOM would be better, but this is a hacky solution for now which does the trick.

jacobSingh:
This isn't the prettiest solution, but it works.

Also commented out some JS in the index.html which was causing an error because that element doesn't exist.

Until the app is moved to a framework, this should do the trick.  Closes #180

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -93,8 +93,11 @@
     <div class="margin-div">
         <h2>Research Report</h2>
         <div id="reportContainer"></div>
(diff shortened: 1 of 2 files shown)


assafelovic (approved):
Great addition, thank you!

PR #209, opened 2023-10-14
ubuntu Dockerfile fix

ElishaKay:
Fixes the scraping capability via Docker for Ubuntu.

Unfortunately doesn't work for mac (crashes on "Docker-compose up --build")

The issue is that the currently Dockerfile doesn't accurately verifying that the Chromium & Chromium Driver Packages installed in the container will be the same

[GPT Discord Thread](https://discord.com/channels/1127851779011391548/1151140679171448934/1161604588978323477
)

Long-term: I would delegate the scraping to a package that specializes in scraping, like:

https://www.npmjs.com/package/node-spider

And create an endpoint which specializes just in scraping an individual page, parsing, and embedding chunks of the text in a Vector DB & returning the text back to the GPT-Agent ([Example here](https://github.com/pinecone-io/pinecone-vercel-starter/blob/main/src/app/api/crawl/seed.ts))

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,17 @@
 FROM python:3.11.4-slim-bullseye as install-browser
 
 RUN apt-get update \
-    && apt-get satisfy -y \
-    "chromium, chromium-driver (>= 115.0)" \
-    && chromium --version && chromedriver --version
+    && apt-get install -y gnupg2 curl unzip
+
+# Add Google Chrome repository and install Chromium and Chromium driver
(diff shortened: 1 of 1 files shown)


assafelovic:
@ElishaKay just to verify, you state that this version doesn't work for Mac?

ElishaKay:
> @ElishaKay just to verify, you state that this version doesn't work for Mac?

Correct. Saw some errors regarding: "google-chrome-stable" not being available in the mac package directory & an error around the "signing key"

PR #228, opened 2023-11-02
Fixes: can't run with firefox as selenium browser

### LINKED ISSUE #227: Bug: can't run with firefox as selenium browser
There are two issues when running GPT researcher with `USE_WEB_BROWSER=firefox`:

1) Firefox is not installed in the dockerfile
2) Accessing GeckoDriver via `webdriver_manager` in each thread quickly overwhelms the github API rate limit of 60 requests per hour - it seems that `webdriver_manager` performs github API requests with each call to `GeckoDriverManager().install()`, even when accessing GeckoDriver from cache, without downloading

I created a pull request that fixes both of these issues: #228

ArtemPt239:
Fixes #227 

- Adds firefox and geckodriver to the dockerfile
- Replaces `GeckoDriverManager().install()` with installing geckodriver in the dockerfile

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,6 +5,12 @@ RUN apt-get update \
     "chromium, chromium-driver (>= 115.0)" \
     && chromium --version && chromedriver --version
 
+RUN apt-get install -y firefox-esr wget \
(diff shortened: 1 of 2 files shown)


assafelovic (approved):
Thanks, this is great!

PR #230, opened 2023-11-04
GPTResearcher Package

gregdrizz:


agent/run.py
```diff
@@ -4,7 +4,8 @@
 from typing import List, Dict
 from fastapi import WebSocket
 from config import check_config_setup
-from agent.research_agent import ResearchAgent
+
+from gptresearcher.gpt_researcher import GPTResearcher
```
assafelovic, line 8: @gregdrizz please add "_" to "gptresearcher" -> "gpt_researcher". Also please change the latter to "agent" under the directory instead of "gpt_researcher".

agent/run.py
assafelovic, line 1: @gregdrizz change this directory from agent to "server" The overall project structure suggestion here is:
[server]
-- run.py
[gpt_researcher]
-- [agent]
[client]
...

agent/run.py
```diff
@@ -34,22 +35,19 @@ async def disconnect(self, websocket: WebSocket):
...
-        report, path = await run_agent(task, report_type, agent, agent_role_prompt, websocket)
+    async def start_streaming(self, task, report_type, websocket):
+        report, path = await run_agent(task, report_type, websocket)
         return report, path
 
 
-async def run_agent(task, report_type, agent, agent_role_prompt, websocket):
-    check_config_setup()
+async def run_agent(task, report_type, websocket):
+    researcher = GPTResearcher()
```
assafelovic, line 44: @gregdrizz a bit confused here but shouldn't GPTResearcher be constructed and initialized with a task and stream type? Then you can call "conduct_researcher" with a specific agent maybe, not sure if that as well should be in construction step.

gptresearcher/actions/web_scrape.py
```diff
@@ -123,14 +129,14 @@ def scrape_text_with_selenium(url: str) -> tuple[WebDriver, str]:
         "firefox": FirefoxOptions,
     }
 
-    options = options_available[CFG.selenium_web_browser]()
```
assafelovic, line 126: @gregdrizz not sure this is who I would go. It's weird passing an object to a util function. The util functions should be redundant of objects such as GPTResearcher. A better approach is to call this functions with the direct params instead of passing an object. So for example the scrap function would get params such as "user_agent", "web_browser", etc. You can also consider writing a higher order method within gpt_researcher that passes the right params and calls this functions. Your call.

gptresearcher/actions/web_search.py
```diff
@@ -7,36 +7,34 @@
 from langchain.utilities import SearxSearchWrapper
 from config import Config
 
-CFG = Config()
 
-
-def web_search(query: str, num_results: int = 4) -> str:
+def web_search(researcher, query: str, num_results: int = 4) -> str:
```
assafelovic, line 11: @gregdrizz same feedback here. don't pass a researcher object, add the relevant params to function. If there are many consider user **kwargs to make the function cleaner. Please keep all util functions generic to gptresearcher object

gptresearcher/context/llm_utils.py
```diff
@@ -3,30 +3,25 @@
...
 
-CFG = Config()
-
-openai.api_key = CFG.openai_api_key
-
 from typing import Optional
-import logging
+
 
 def create_chat_completion(
```
assafelovic, line 17: @gregdrizz you can see here an example of passing many params instead of an object

gptresearcher/context/llm_utils.py
```diff
@@ -98,7 +98,7 @@ async def stream_response(model, messages, temperature, max_tokens, websocket):
     return response
 
 
-def choose_agent(task: str) -> dict:
+def choose_agent(researcher, task: str) -> dict:
```
assafelovic, line 101: @gregdrizz same here. please remove all researcher objects from generic functions. Pass the relevant params such as "model"

gptresearcher/context/research_context.py
```diff
@@ -5,44 +5,40 @@
...
-
-
-class ResearchAgent:
-    def __init__(self, question, agent, agent_role_prompt, websocket=None):
-        """ Initializes the research assistant with the given question.
-        Args: question (str): The question to research
-        Returns: None
-        """
-
+class ResearchContext:
```
assafelovic, line 22: @gregdrizz as discussed it's better consider it as "RetrieverAgent" than Context. Makes more sense. That way in theory developers or us can initiate multiple RetrieverAgents based on various sources/use cases/domains, etc. And again, instead of passing a researcher object, not sure why not to pass the agent_role, agent, etc as originally planned? It would make this object be more abstract and customizable for various sub research agents.

gptresearcher/gpt_researcher.py
```diff
@@ -0,0 +1,109 @@
...
+        for query in search_queries:
+            research_result = await research_context.run_search_summary(query)
+            research_context.research_summary += f"{research_result}\n\n"
+
+        report, path = await research_context.write_report(report_type, websocket)
+
+        return report, path
+
+
+import asyncio
```
assafelovic, line 97: @gregdrizz please move all imports to top of file, also not sure both of these should be in same file (main + class object). Would love to understand the logic behind it. I'd consider having a separate file for the actual GPTResearcher object and the main one. Even if that's how it today, it wasn't intended and should be refactored.

gptresearcher/gpt_researcher.py
```diff
@@ -0,0 +1,109 @@
+from typing import Optional
+import os
+
+import openai
+
+from gptresearcher.context.research_context import ResearchContext
+from gptresearcher.utils.setup_check import check_agent_setup, check_openai_api_key
+
+
+class GPTResearcher:
```
assafelovic, line 10: @gregdrizz please see how to pass config file string as alternative to all these params. Consider rewriting our config file to be a JSON structure that can be imported into the class. Also consider modifying all these params to use **kwargs for a cleaner use. Please see examples for such best practices in other open sources or in general.

See here very broad example:
class MyClass(object):
    def __init__(self, **kwargs):
        for key, value in kwargs.items():
            setattr(self, key, value)

    @classmethod
    def from_json(cls, file_path):
        with open(file_path, 'r') as f:
            config = json.load(f)
        return cls(**config)
        
obj1 = MyClass(param1=value1, param2=value2, ...)
obj2 = MyClass.from_json('path/to/config.json')

Please make sure that optional params (such as debug_mode, user_agent, etc), I can think of 80% of these params being optional, that they are considered in **kwargs and dealt with a default value in __init__.

gpt_researcher/master/agent.py
```diff
@@ -68,17 +84,13 @@ async def run_sub_query(self, sub_query):
...
         retriever = self.retriever(sub_query)
-        urls = retriever.search()
-        urls_to_scrape = []
-        for url in urls:
-            if url not in self.visited_urls:
-                await self.stream_output("logs", f"✅ Adding source url to research: {url}\n")
-                self.visited_urls.add(url)
-                urls_to_scrape.append(url)
+        search_results = retriever.search()
+        new_search_urls = await self.get_new_urls([url.get("href") for url in search_results])
```
gregdrizz, line 88: Seems like this line works only for tavily, duckduckgo's response is just a list of url's, maybe we should add a new_search_urls method to the retriever class.

PR #237, opened 2023-11-07
Fix Google SerpAPI bug (Update the SerpAPI to Latest Version)

ray-ruisun:
The google updated serp api version, the old one will trigger problems.


ray-ruisun:
https://serpapi.com/ in this way, users could use a free subscription to test GPTResearcher with googleSerp search engine.

assafelovic (approved):
Thank you!

PR #261, opened 2023-11-21
Fix serper retriever

reasonmethis:
Background: [PR 237](https://github.com/assafelovic/gpt-researcher/pull/237) introduced a mix-up, it replaced the correct code for the Google Serper retriever (serper.dev) with code for SerpApi (serpai.com). It looks like the author of the PR may have confused one for the other and thought that the existing correct code was a bug.

This PR:
- brings back the correct code (with small improvements)
- adds the SerApi code from PR 237 as a separate retriever (although that code appears to have an issue, as I noted in the code itself, thus for now I didn't fully enable that retriever)
- resolves a couple of naming issues (e.g. 'serp' vs 'googleSerp' as the name of the retriever)

### DIFF
--- a/gpt_researcher/master/functions.py
+++ b/gpt_researcher/master/functions.py
@@ -25,7 +25,7 @@ def get_retriever(retriever):
         case "searx":
             from gpt_researcher.retrievers import SearxSearch
             retriever = SearxSearch
-        case "serp":
+        case "googleSerp":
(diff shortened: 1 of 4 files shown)


assafelovic (approved):

assafelovic:
Thank you @reasonmethis super helpful!

PR #262, opened 2023-11-21
Added Config UI

Ethan-Otto:
Added UI for controlling the config. I made the config.json in the root folder the default config, and the UI can update it directly. Allows much faster testing of different parameters especially when using docker.
Also added descriptions of each variable

image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,4 +31,4 @@ USER gpt-researcher
 COPY --chown=gpt-researcher:gpt-researcher ./ ./
 
 EXPOSE 8000
-CMD ["uvicorn", "main:app", "--host", "0.0.0.0", "--port", "8000"]
+CMD ["uvicorn", "main:app", "--host", "0.0.0.0", "--port", "8000", "--reload",  "--log-level", "debug"]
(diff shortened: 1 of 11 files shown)


assafelovic:
@Ethan-Otto this is great! Can you please refine and include on the following params:

Report format, total words, search engine, llm provider and max iterations? The idea is to expose only the report related params. Thanks!

Ethan-Otto:
@assafelovic So hide everything else?

Ethan-Otto:
@assafelovic so hide everything else?

assafelovic:
Not to hide but remove, I think some of these config params should be hidden and not modified since it may confuse or break the UX/UI. Lmk if you can, if not I can do it later in the week

notvicent3 (commented):
Didn't check it running atm, but as I can see in the code seems pretty nice.

Ethan-Otto:
@assafelovic Added a filter to control what's displayed/updated
image

assafelovic:
Unfortunately this PR is stale and consists of too many conflicts. Also we've shifted focus to a new NextJS app. Closing this.

PR #284, opened 2023-12-03
Add Bing Search Retriever

norrisp90:
Adding BingSearch Bing Web Search API retriever. Requires ["BING_API_KEY"] env variable set to work.


assafelovic (approved):
Looking good!

PR #285, opened 2023-12-03
Adding Azure OpenAI Model Support

norrisp90:
Adding supoprt for Azure OpenAI model deployments.

Note: These are breaking changes that mean only Azure OpenAI models will working using them. To merge these in correctly logic will be required in both the memory embeddings and LLM functions so that when self.llm_provider = "AzureChatOpenAI" the correct Azure version of the code is used.


assafelovic:
This can be used manually by any developer but not as default. Thanks for suggesting!

PR #322, opened 2024-01-04
Fix conflicting openai package versions in requirements.txt

saeedghadiri:


cryolite-ai:
@assafelovic This helpful PR from @saeedghadiri should be a really trivial merge frankly - currently the error is breaking for installation for any new user doing a manual install (which looks bad/sad); might be worth rummaging back to find the point at which duplicates happened..

Also, maybe worth removing one of the duplicate playwright==1.40.0  requirement entry in the requirements.txt (both playwright package specifications are the same, so it doesn't cause installation to break like the conflicting openai package versions do, but there's clearly been an issue with merging together some requirements.txt files at some point..) 

Thanks!

assafelovic:
@saeedghadiri @cryolite-ai thanks for raising this. Was done due to PR bot. Pushed a quick fix so closing this but super appreciate catching it!

PR #350, opened 2024-02-02
build: update compatibility of third-party packages

Undertone0809:
**PR description**f

optimize compatibility of third-party packages.

**Problem description**

My project uses poetry as a package management tool, which locks all versions of third-party libraries when running `poetry lock`. Lock requirements.txt third-party packages version will make users using Poetry very unfriendly, and fixing a certain version can easily lead to version conflicts with third-party libraries that developers are currently using.

The dependencies I have modified are all non breaking versions.

image

**Subsequent operations**

It is highly recommended to release a new pypi version after this PR to improve the compatibility of third-party libraries.


Undertone0809:
@assafelovic Can you review it?

assafelovic:
@Undertone0809 I've removed strict versions completely with latest version so it should work better now.

PR #365, opened 2024-02-23
A new url scraper!

proy9714:
Introduced **newspaper3k** scraper as an alternate scraper to BeautifulSoup. **Newspaper3k** provides a quick and easy way to scrape online urls. Also provides functionality for scraping images, which may be explored in the future.

Introduced a new function scrape_url_with_newspaper in scraper.py which uses the newspaper3k functionality and returns the scraped text.

A new instance variable "scraper" has been introduced which controls which scraper is to take action "bs" -> beautifulsoup or newspaper3k.

Also modified _**requirements.txt**_ and added the newspaper3k library.


assafelovic:
This is super interesting addition thank you @proy9714

assafelovic:
@proy9714 would you mind making a slight change and move the "bs" option to config.py and config.json? Then extract it within the `gpt_researcher/master/functions.py` file and pass to `content field`. 

you can add a field for scraper like so (for example in config.py):
`
self.scraper="bs"
`
and then in the `scrape_urls` method in `gpt_researcher/master/functions.py` pass it with cfg.scraper

proy9714:
@assafelovic Here are the requested changes! Let me know if anything else needs changing...

assafelovic (approved):
Thanks great addition!

assafelovic:
Thanks @proy9714 this is great!

PR #375, opened 2024-03-05
Add support for various embedding provider

jimmylin0979:
Introduce support for additional embedding models, such as Hugging Face, MixTRAL, Ollama, etc., allowing users to select their preferred model in the config.json file simply. Any model supported by [langchain embedding](https://python.langchain.com/docs/integrations/text_embedding) can be added easily.

Particularly beneficial for users who do not have access to the OpenAI API, like me XD.

### DIFF
--- a/config.json
+++ b/config.json
@@ -4,6 +4,7 @@
             "allow_downloads": false,
             "selenium_web_browser": "chrome",
             "search_api": "tavily",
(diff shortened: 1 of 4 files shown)


assafelovic:
Great addition thank you @jimmylin0979

PR #419, opened 2024-04-01
Removes asyncio as it is in the std lib

sockthedev:
Has been available in the [std lib](https://docs.python.org/3.10/library/asyncio.html) for a while now.

Including it as a dependency causes issues in my runtime within AWS Lambda.


assafelovic:
Thanks @sockthedev

assafelovic (approved):

PR #428, opened 2024-04-04
Quick fixes

proy9714:
1. Included parent query during search question generation to retain context as requested by @assafelovic 
2. A quick fix to the example file according to latest release.


assafelovic (approved):
100%!

Issue #427, opened 2024-04-04
AzureOpenAI Configuration Error


deb1305:
I am trying to configure azure openai llm. After setting the config.py file, I am getting below error:

ERROR:    Exception in ASGI application

Can someone share the config.py with respect to azure openai llm

assafelovic:
@norrisp90 can you give a hand here? I don't have an Azure account

norrisp90:
@assafelovic Same issue as #422 . Will work on it in #422 .

deb1305:
that passed through, but now i am getting below error 
raise DuckDuckGoSearchException(f"_get_url() {url} {type(ex).__name__}: {ex}")
duckduckgo_search.exceptions.DuckDuckGoSearchException: _get_url() https://duckduckgo.com RequestsError: Failed to perform, ErrCode: 60, Reason: 'SSL certificate problem: unable to get local issuer certificate'. This may be a libcurl error, See https://curl.se/libcurl/c/libcurl-errors.html first for more details.

Note: I am using Tavily api key

norrisp90:
> that passed through, but now i am getting below error raise DuckDuckGoSearchException(f"_get_url() {url} {type(ex).**name**}: {ex}") duckduckgo_search.exceptions.DuckDuckGoSearchException: _get_url() https://duckduckgo.com RequestsError: Failed to perform, ErrCode: 60, Reason: 'SSL certificate problem: unable to get local issuer certificate'. This may be a libcurl error, See https://curl.se/libcurl/c/libcurl-errors.html first for more details.
> 
> Note: I am using Tavily api key

@deb1305 This looks like a seperate issue. Suggest you open new issue for it.

assafelovic:
Fixe here: https://github.com/assafelovic/gpt-researcher/pull/432

Issue #422, opened 2024-04-02
AzureOpenAI - error generating subtopics in detailed mode


Xtrah:
To use AzureOpenAI, I have the following `.env`:
```
EMBEDDING_PROVIDER=azureopenai
LLM_PROVIDER=azureopenai
SMART_LLM_MODEL=gpt-4-0125-preview
AZURE_OPENAI_API_KEY=REDACTED
AZURE_OPENAI_ENDPOINT=https://REDACTED.openai.azure.com/
OPENAI_API_VERSION=2024-02-01
AZURE_EMBEDDING_MODEL=text-embedding-3-large
OPENAI_API_TYPE="azure"
TAVILY_API_KEY=REDACTED
```

Running short and fast mode works fine. 

However, detailed mode stops working when it starts generating subtopics:

```
🤔 Generating subtopics...

🤖 Calling gpt-4-0125-preview...

Exception in parsing subtopics :  1 validation error for ChatOpenAI
__root__
  Did not find openai_api_key, please add an environment variable `OPENAI_API_KEY` which contains it, or pass `openai_api_key` as a named parameter. (type=value_error)
📋Subtopics: []
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/site-packages/uvicorn/protocols/websockets/wsproto_impl.py", line 233, in run_asgi
    result = await self.app(self.scope, self.receive, self.send)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/uvicorn/middleware/proxy_headers.py", line 69, in __call__
    return await self.app(scope, receive, send)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/fastapi/applications.py", line 1054, in __call__
    await super().__call__(scope, receive, send)
  File "/usr/local/lib/python3.11/site-packages/starlette/applications.py", line 123, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/usr/local/lib/python3.11/site-packages/starlette/middleware/errors.py", line 151, in __call__
    await self.app(scope, receive, send)
  File "/usr/local/lib/python3.11/site-packages/starlette/middleware/exceptions.py", line 65, in __call__
[...]

PR #432, opened 2024-04-04
Hotfix for #422 & # 427 : LLM provider = azureopenai when using detailed report mode

norrisp90:
Hotfix for #422 & #427  Detailed report azureopenai support plus azureopenai LLM settings documentation improvement.

### DIFF
--- a/gpt_researcher/llm_provider/azureopenai/azureopenai.py
+++ b/gpt_researcher/llm_provider/azureopenai/azureopenai.py
@@ -4,13 +4,18 @@ from colorama import Fore, Style
 from langchain_openai import AzureChatOpenAI
 
 '''
-Please note: Needs additional env vars such as: 
-    AZURE_OPENAI_ENDPOINT  e.g. https://xxxx.openai.azure.com/", 
-    OPENAI_API_VERSION, 
-    OPENAI_API_TYPE
+Please note:
+Needs additional env vars such as: 
+    AZURE_OPENAI_ENDPOINT  e.g. https://xxxx.openai.azure.com/",
+    AZURE_OPENAI_API_KEY e.g "xxxxxxxxxxxxxxxxxxxxx",
+    OPENAI_API_VERSION, e.g. "2024-03-01-preview" but needs to updated over time as API verison updates,
+    AZURE_EMBEDDING_MODEL e.g. "ada2" The Azure OpenAI embedding model deployment name.
 
-Note new entry in config.py to specify the Azure OpenAI embedding model name:
-self.azure_embedding_model = os.getenv('AZURE_EMBEDDING_MODEL', "INSERT_EMBEDDIGN_MODEL_DEPLOYMENT_NAME")
+config.py settings for Azure OpenAI should look like:
+    self.embedding_provider = os.getenv('EMBEDDING_PROVIDER', 'azureopenai')
+    self.llm_provider = os.getenv('LLM_PROVIDER', "azureopenai")
+    self.fast_llm_model = os.getenv('FAST_LLM_MODEL', "gpt-3.5-turbo-16k") #Deployment name of your GPT3.5T model as per azure OpenAI studio deployment section
+    self.smart_llm_model = os.getenv('SMART_LLM_MODEL', "gpt4")  #Deployment name of your GPT4 1106-Preview+ (GPT4T) model as per azure OpenAI studio deployment section
 '''
 class AzureOpenAIProvider:
 
--- a/gpt_researcher/utils/llm.py
+++ b/gpt_researcher/utils/llm.py
@@ -123,7 +123,13 @@ async def construct_subtopics(task: str, data: str, config, subtopics: list = []
 
         print(f"\n🤖 Calling {config.smart_llm_model}...\n")
 
-        model = ChatOpenAI(model=config.smart_llm_model)
+        if config.llm_provider == "openai":
+            model = ChatOpenAI(model=config.smart_llm_model)
+        elif config.llm_provider == "azureopenai":
+            from langchain_openai import AzureChatOpenAI
+            model = AzureChatOpenAI(model=config.smart_llm_model)
+        else:
+            return []
 
         chain = prompt | model | parser
 


gpt_researcher/llm_provider/azureopenai/azureopenai.py
```diff
@@ -4,13 +4,18 @@
...
-Please note: Needs additional env vars such as: 
-    AZURE_OPENAI_ENDPOINT  e.g. https://xxxx.openai.azure.com/", 
-    OPENAI_API_VERSION, 
-    OPENAI_API_TYPE
+Please note:
+Needs additional env vars such as: 
+    AZURE_OPENAI_ENDPOINT  e.g. https://xxxx.openai.azure.com/",
+    AZURE_OPENAI_API_KEY e.g "xxxxxxxxxxxxxxxxxxxxx",
+    OPENAI_API_VERSION, e.g. "2024-03-01-preview" but needs to updated over time as API verison updates,
+    AZURE_EMBEDDING_MODEL e.g. "ada2" The Azure OpenAI embedding model deployment name.
```
Xtrah, line 12:
```suggestion
Please note:
AzureOpenAI requires additional env vars such as: 
    EMBEDDING_PROVIDER="azureopenai"
    LLM_PROVIDER="azureopenai"
    SMART_LLM_MODEL="{LLM_DEPLOYMENT_NAME}"
    FAST_LLM_MODEL="{LLM_DEPLOYMENT_NAME}"
    AZURE_OPENAI_ENDPOINT="https://{ENDPOINT}.openai.azure.com/",
    AZURE_OPENAI_API_KEY="{API_KEY}",
    OPENAI_API_VERSION="2024-02-01",
    AZURE_EMBEDDING_MODEL="{EMBEDDING_MODEL_DEPLOYMENT_NAME}"
    OPENAI_API_TYPE="azure"
```

gpt_researcher/llm_provider/azureopenai/azureopenai.py
```diff
@@ -4,13 +4,18 @@
...
 
-Note new entry in config.py to specify the Azure OpenAI embedding model name:
-self.azure_embedding_model = os.getenv('AZURE_EMBEDDING_MODEL', "INSERT_EMBEDDIGN_MODEL_DEPLOYMENT_NAME")
+config.py settings for Azure OpenAI should look like:
+    self.embedding_provider = os.getenv('EMBEDDING_PROVIDER', 'azureopenai')
+    self.llm_provider = os.getenv('LLM_PROVIDER', "azureopenai")
+    self.fast_llm_model = os.getenv('FAST_LLM_MODEL', "gpt-3.5-turbo-16k") #Deployment name of your GPT3.5T model as per azure OpenAI studio deployment section
+    self.smart_llm_model = os.getenv('SMART_LLM_MODEL', "gpt4")  #Deployment name of your GPT4 1106-Preview+ (GPT4T) model as per azure OpenAI studio deployment section
```
Xtrah, line 18:
```suggestion
```

Xtrah (commented):
Isn't it cleaner to just use a `.env` file considering config.py is made to get environment variables?

Some suggestions below based on that, if you agree.

gpt_researcher/llm_provider/azureopenai/azureopenai.py
norrisp90: In theory but the config.py pattern is what is currently used and we need to support multiple model types, if you move these Azure OpenAI ones to .env then you are breaking the current design? To implement this properly i think you would have to remove the references to the config.py file values where they are used throughoput the codebase and replace with stright env. var calls?

norrisp90:
> Isn't it cleaner to just use a `.env` file considering config.py is made to get environment variables?
> 
> Some suggestions below based on that, if you agree.

In theory but the config.py pattern is what is currently used, if you move these to .env then you are braking the current design. To implement this properly i think you would have to remove the refences to teh config.py file values where they are used throughoput the codebase and replace with stright env. var calls?

> Isn't it cleaner to just use a `.env` file considering config.py is made to get environment variables?
> 
> Some suggestions below based on that, if you agree.

In theory but the config.py pattern is what is currently used and we need to support multiple model types, if you move these Azure OpenAI ones to .env then you are breaking the current design? To implement this properly i think you would have to remove the references to the config.py file values where they are used throughoput the codebase and replace with stright env. var calls? @Xtrah

Xtrah:
> > Isn't it cleaner to just use a `.env` file considering config.py is made to get environment variables?
> > Some suggestions below based on that, if you agree.
> 
> In theory but the config.py pattern is what is currently used, if you move these to .env then you are breakinh the current design. To implement this properly i think you would have to remove the references to the config.py file values where they are used throughoput the codebase and replace with stright env. var calls? @Xtrah

Isn't it supported by the current design, considering you are using:

```
self.smart_llm_model = os.getenv('SMART_LLM_MODEL', "gpt-4-1106-preview")
```

Where the optional second argument is the default. So you can either specify changes in .env or you can change the default, both with the same result?

norrisp90:
> > > Isn't it cleaner to just use a `.env` file considering config.py is made to get environment variables?
> > > Some suggestions below based on that, if you agree.
> > 
> > 
> > In theory but the config.py pattern is what is currently used, if you move these to .env then you are breakinh the current design. To implement this properly i think you would have to remove the references to the config.py file values where they are used throughoput the codebase and replace with stright env. var calls? @Xtrah
> 
> Isn't it supported by the current design, considering you are using:
> 
> ```
> self.smart_llm_model = os.getenv('SMART_LLM_MODEL', "gpt-4-1106-preview")
> ```
> 
> Where the `os.getenv` optional second argument is the default. So you can either specify changes in .env or you can change the default, both with the same result? That's why I was thinking it is cleaner to just do changes in one file that is already dynamic, rather than the code.

Not really. If you place something in env var it will only work if it's called directly as env in the code. Calls like config.llm_provider are used extensively also which would all need to be replaced if everything was to move to env variables.

Your idea on defaults also doesn't consider when llm_provider type is set to another model provider other than Azure OpenAI which work differently and would need other defaults or none at all etc.

deb1305 (commented):
this fix is working

Xtrah (approved):
@norrisp90 considering config.py uses os.getenv for all variables, and we are only calling on config.variable in the code, it will work both by setting new defaults in config.py and setting a variable in .env. Setting the variable in .env will overwrite whatever is set in config.py. An advantage of using .env allows to e.g. host a Docker image on GHCR or Docker Hub that is configurable right out of the box with just .env and no code changes, opening up for more users.

Thank you for fixing this so quickly, I have confirmed [it works great](https://gist.github.com/Xtrah/167dbd3ab0eb5f38fa39006dc27d2c12)!

assafelovic:
Thank you @norrisp90 !

PR #444, opened 2024-04-14
fix: add lxml with html_clean extra to pyproject.toml

ldorigo:
Currently, cannot start project if installed with `poetry install`


assafelovic (approved):

assafelovic:
Thank you @ldorigo

PR #462, opened 2024-05-02
langgraph multi-agents feedback

ElishaKay:
Some food for thought based on playing with the multi-agents feature:

a) Had to some errors in the CLI regarding missing dependencies - also important to stress in the Docs that the user should use Python 3.10 and up

For example:

`python3.10 -m pip install python-dotenv`
`python3.10 main.py`

b) I feel that the LangSmith UI features here could be a powerful tool for:
 - enabling users to visualize and inspect the backend data flow
 - quality assurance debugging - i.e. where can the input or output use improvement

Attaching here a screenshot from https://smith.langchain.com
A big added value here would be:
 - customizing the name of each step as it appears within the LangSmith GUI.
 For example: "Step 1: Auto-Generate Agent Prompt"...
  - Perhaps adding a tag per question for easy filtering within the LangSmith GUI.

Screen Shot 2024-05-02 at 22 32 12


assafelovic (approved):
Thanks this is great!

PR #463, opened 2024-05-03
fix: Use total_words parameters in the prompts

diegonobre:
While reviewing the prompts, I noticed that on a few prompts the parameter `total_words` is not being used.

Let me know if this is ok to merge or if it requires additional changes


assafelovic (approved):
This is great thank you!

PR #497, opened 2024-05-13
Update trobuleshooting doc to handle installation bugs related to lib…

satishjasthi:
Update trobuleshooting doc to handle installation bugs related to lib pango and glib in MacOS and linux


assafelovic (approved):
Thank you!

PR #501, opened 2024-05-16
WIP: add langgraph deploy

hwchase17:
You can now deploy this with:

```
pip install -U langgraph-cli
langgraph up -c langgraph.json
```

See here for instructions on how to use once up: https://github.com/langchain-ai/langgraph-example

Is WIP because is lacking good instructions, and also I don't love how task is passed in, once at init and another at run time

### DIFF
--- a/docs/docs/gpt-researcher/llms.md
+++ b/docs/docs/gpt-researcher/llms.md
@@ -10,6 +10,39 @@ Below you can find how to configure the various supported LLMs.
 
 ## Groq
 
+GroqCloud provides advanced AI hardware and software solutions designed to deliver amazingly fast AI inference performance.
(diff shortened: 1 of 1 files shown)


multi_agents/agents/utils/file_formats.py
```diff
@@ -2,7 +2,6 @@
 import urllib
 import uuid
 import mistune
-from md2pdf.core import md2pdf
```
assafelovic, line 5: @hwchase17 what's the reason for this?

multi_agents/agent.py
```diff
@@ -0,0 +1,16 @@
+from agents import ChiefEditorAgent
```
assafelovic, line 1: @hwchase17 wouldn't it be better to open the task.json file here?

PR #503, opened 2024-05-16
automated tests

ElishaKay:
Run tests with:

`pytest -s test_gpt_researcher.py`

### DIFF
--- a/test_gpt_researcher.py
+++ b/test_gpt_researcher.py
@@ -1,0 +1,50 @@
+import os
+import asyncio
+import pytest
+from gpt_researcher import GPTResearcher
+
+# Define the report types to test
+report_types = [
+    "research_report",
+    "custom_report",
+    "subtopic_report",
+    "summary_report",
+    "detailed_report",
+    "quick_report"
+]
+
+# Define a common query and sources for testing
+query = "What are the latest advancements in AI?"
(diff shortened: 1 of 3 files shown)


test_gpt_researcher.py
```diff
@@ -0,0 +1,50 @@
...
+    # Conduct research and write the report
+    await researcher.conduct_research()
+    report = await researcher.write_report()
+    
+    # Define the expected output filenames
+    pdf_filename = os.path.join(output_dir, f"{report_type}.pdf")
+    docx_filename = os.path.join(output_dir, f"{report_type}.docx")
+    
+    # Check if the PDF and DOCX files are created
+    # assert os.path.exists(pdf_filename), f"PDF file not found for report type: {report_type}"
```
assafelovic, line 42: @ElishaKay why is this commented? Is this test final?

ElishaKay: For those 6 use cases in the commit message above, correct, the tests are final and pass.

I need to make a change in agent.py for those tests to pass (`initialize report = ""` at top of the function. If that doesn't break anything feel free to merge. Else we can continue writing tests for more use cases here.

Food.for thought:

 - What are the potential parameters and values that can be passed to the GPTResearcher Class?
 - For end to end tests (fronted to backend and back), we'd be better positioned leveraging a React Testing Library

assafelovic: Just got me thinking that GPT Researcher should be able to expose the writing functions

PR #509, opened 2024-05-20
remove DOC_Path from docker-compose - rely on .env instead

ElishaKay:
Fix 'docker-compose up --build' command for docs

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,7 +6,6 @@ services:
     environment: 
       OPENAI_API_KEY: ${OPENAI_API_KEY}
       TAVILY_API_KEY: ${TAVILY_API_KEY}
-      DOC_PATH: "=./my-docs"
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -6,7 +6,6 @@ services:
     environment: 
       OPENAI_API_KEY: ${OPENAI_API_KEY}
       TAVILY_API_KEY: ${TAVILY_API_KEY}
-      DOC_PATH: "=./my-docs"
```
assafelovic, line 9: Shouldn't it be DOC_PATH: ${DOC_PATH} ?

ElishaKay: No need. Relying on .env is preferable

Path variables get tricky and run into edge cases when dealing with containers.

Better to play it safe and work with what we know works until.peoven otherwise

assafelovic (approved):

Issue #511, opened 2024-05-20
Incorrect setting of alternative base_url for ChatOpenAI


mmashnev:
I encountered an error while attempting to configure an alternative OpenAI base URL via the environment parameter `OPENAI_BASE_URL`. The error log is attached below for reference.

Error Log:

```
2024-05-20 19:34:12 INFO:     ('172.18.0.1', 56588) - "WebSocket /ws" [accepted]
2024-05-20 19:34:12 INFO:     connection open
2024-05-20 19:34:19 ERROR:    Exception in ASGI application
2024-05-20 19:34:19 Traceback (most recent call last):
2024-05-20 19:34:19   File "/usr/local/lib/python3.11/site-packages/uvicorn/protocols/websockets/websockets_impl.py", line 240, in run_asgi
2024-05-20 19:34:19     result = await self.app(self.scope, self.asgi_receive, self.asgi_send)
2024-05-20 19:34:19              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2024-05-20 19:34:19   File "/usr/local/lib/python3.11/site-packages/uvicorn/middleware/proxy_headers.py", line 69, in __call__
2024-05-20 19:34:19     return await self.app(scope, receive, send)
2024-05-20 19:34:19            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2024-05-20 19:34:19   File "/usr/local/lib/python3.11/site-packages/fastapi/applications.py", line 1054, in __call__
2024-05-20 19:34:19     await super().__call__(scope, receive, send)
2024-05-20 19:34:19   File "/usr/local/lib/python3.11/site-packages/starlette/applications.py", line 123, in __call__
2024-05-20 19:34:19     await self.middleware_stack(scope, receive, send)
2024-05-20 19:34:19   File "/usr/local/lib/python3.11/site-packages/starlette/middleware/errors.py", line 151, in __call__
2024-05-20 19:34:19     await self.app(scope, receive, send)
2024-05-20 19:34:19   File "/usr/local/lib/python3.11/site-packages/starlette/middleware/exceptions.py", line 65, in __call__
2024-05-20 19:34:19     await wrap_app_handling_exceptions(self.app, conn)(scope, receive, send)
[...]

PR #512, opened 2024-05-20
Fix Pydantic validation of base_url assignment for ChatOpenAI model

mmashnev:
Pydantic alias 'base_url' for the field 'openai_api_base' is not working as expected for assignments outside the constructor, so changed the alias to the original field name for alternative OpenAI base_url.

Fix for the issue #511


assafelovic (approved):
Perfect, thank you!

PR #543, opened 2024-05-26
Adding Arxiv Search

Ayman-Ameen:
- Adding archive search for more accurate and scientific results.
- Fixing some issues in importing libraries in Apple Silicon.


backend/utils.py
```diff
@@ -1,5 +1,15 @@
 import aiofiles
 import urllib
+import uuid
+try:
```
assafelovic, line 4: @Ayman-Ameen can you explain this? We've already added try catch for importing this libraries within the function calls. Did you merge latest master?

gpt_researcher/master/functions.py
```diff
@@ -40,6 +40,9 @@ def get_retriever(retriever):
         case "bing":
             from gpt_researcher.retrievers import BingSearch
             retriever = BingSearch
+        case "ArxivSearch":
```
assafelovic, line 43: Change to "arxiv". We'd like lower case and simple calls

assafelovic:
Stale PR with no response

PR #546, opened 2024-05-27
RFC: generic llm provider

baskaryan:
Proposal for a more generic LLM provider wrapper to make it easier to add support for arbitrary providers


IA-Programming:
why this is still in draft what else is need to be done?
maybe can be something that I can do

assafelovic (approved):
This is great!

PR #556, opened 2024-05-31
Read the correct variable name in together.py

fei0810:
Read the correct variable name, "MISTRAL_API_KEY" to "TOGETHER_API_KEY"


assafelovic (approved):
Thanks good catch!

PR #559, opened 2024-06-01
Fix[readme]: Page Not Found

gkhngyk:
Replace the URL giving **Page Not Found** with the correct URL.

[Incorrect URL](https://docs.gptr.dev/docs/gpt-researcher/agent_frameworks)

[Correct URL](https://docs.gptr.dev/docs/gpt-researcher/langgraph)


assafelovic (approved):
Good catch!

PR #596, opened 2024-06-13
use repair_json() on output of sub_queries to stop exceptions on local models

pfrench42:
Many of my ollama models are outputting slightly malformed json which was causing exceptions during searches.

This was happening on over 50% of queries, but this small patch resolved


assafelovic (approved):
Great addition thank you!

PR #597, opened 2024-06-14
fix: similarity_threshold config

gkhngyk:
-fix: duplicated and wrong self.fast_token_limit


assafelovic (approved):
Great call thank you!

PR #599, opened 2024-06-15
Lazily import langchain-huggingface to avoid obligatory install

ldorigo:
This could also be done for other providers, but huggingface is particularly heavy to install due to dependencies on torch &more, so kept it to that for now.


assafelovic (approved):
Thank you great addition

PR #610, opened 2024-06-18
extract_headers() was throwing an exception when it was examining a l…

pfrench42:
extract_headers() was throwing an exception when it was examining a line which read "hr".
This fix ensures only hN lines which have digits are parsed

```
> /gpt_researcher/master/actions.py(331)extract_headers()
-> level = int(line[2])  # Extract header level
(Pdb) p line
'<hr />'
(Pdb) 
```


pfrench42:
To be clear, there were <hr > markers in the output which was causing line 331 to throw an exception and abort before research completion.

assafelovic (approved):
great addition!

PR #616, opened 2024-06-20
Implement more graceful JSON handling on choose_agent in master/actions.py

AADaoud:
Reason: When I was testing the limits of this project I noticed that passing off any instructions through the agent as a query broke the choose_agent function as the response contained non-JSON formatted text.

Suggestion: I have added two new async functions to handle different JSON errors using the already imported json_repair library and regex as a last resort.

Importance: This contribution is important because it makes the user experience remain relevant to the research task even when the user (who might not be technically adept) decides to pass instructions in the prompt.

### DIFF
--- a/gpt_researcher/master/actions.py
+++ b/gpt_researcher/master/actions.py
@@ -2,6 +2,7 @@ import asyncio
 import json
 
 import markdown
+import re
 
 from gpt_researcher.master.prompts import *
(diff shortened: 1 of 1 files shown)


assafelovic (approved):
This is great!

assafelovic:
Thank you for this improvement!

assafelovic:
Thank you for this improvement!

PR #674, opened 2024-07-15
Fix tone issue related to the default constructor of GPTResearcher

0x11c11e:
The constructor of GPTResearcher was unable to pick up the default tone when none was provided, causing an issue when calling GPTResearcher without specifying a tone. This change will fix this bug.


assafelovic (approved):
Thank you!

PR #682, opened 2024-07-18
Added gptr-logo to frontend/static

muhammedsaidckr:
Added gptr-logo.png to frontend/static folder. When uvicorn was started with main:app, the project gptr-logo was not displayed and this way the problem is eliminated.


assafelovic (approved):
Perfect, thank you!

PR #679, opened 2024-07-17
Sanitize Directory Names to Ensure Compatibility with Windows

0x11c11e:
**Description:**
This PR introduces a function to sanitize directory names to ensure compatibility with Windows operating systems. The primary change is in the ChiefEditorAgent class, where the directory name for storing output files is now sanitized to replace invalid characters.

**Changes Made:**
- Added sanitize_filename function to replace invalid characters in directory names.
- Updated ChiefEditorAgent to use the sanitize_filename function when creating output directories.

**Background:**
The project was functioning correctly on Linux but encountered errors on Windows due to invalid characters in directory names (e.g., ?, *, |). This update addresses those issues by sanitizing the directory names, making the project cross-platform compatible.

PR #683, opened 2024-07-18
Fixes: Default retriever, sanitizing

ruzicka02:
### Default retriever
In the previous version, the default case in get_retriever() meant
that setting retriever from config (as opposed to the header) was
ignored, and instead, the default (tavily) retriever was chosen.
Added new function to get the default retriever.

### Corrected sanitizing
Sanitizing function replaces the slashes in the path, which is
most likely undesired. Discussed in 
https://github.com/assafelovic/gpt-researcher/pull/679#issuecomment-2235946414

### DIFF
--- a/gpt_researcher/master/actions.py
+++ b/gpt_researcher/master/actions.py
@@ -64,10 +64,13 @@ def get_retriever(retriever):
             retriever = CustomRetriever
 
         case _:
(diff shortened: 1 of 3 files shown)


assafelovic (approved):
Love this, thank you!

PR #700, opened 2024-07-23
Source url research feature added and Fixes report_source and report_type value on multi agents side

muhammedsaidckr:
1. In this way, source_urls to be researched can be added and research is done only through these source_urls
2. Frontend changes have been applied. If report_source is selected as sources, the input field opens and source_urls can be added.
image

On multi agent side 

report_type and report_source have been replaced with names from the parameter as reportType and reportSource

### DIFF
--- a/backend/server.py
+++ b/backend/server.py
@@ -77,6 +77,7 @@ async def websocket_endpoint(websocket: WebSocket):
                 json_data = json.loads(data[6:])
                 task = json_data.get("task")
                 report_type = json_data.get("report_type")
+                source_urls = json_data.get("source_urls")
                 tone = json_data.get("tone")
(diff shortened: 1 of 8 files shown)


FarVision2:
Yes, this is going to be necessary for me to use the tool. I have not used it in a while so I've recently come back. Here is a list of the sources for my last two multi-agent queries.  

forbes
democracyforward
wikipedia
snopes
bbc
cbsnews
usatoday
nytimes

See anything wrong with this picture? You can have the disposition of the prompt set to be as neutral as possible but you're not going to get a good research project with all biased sources. Whichever way you lean. Is there a source list in the code somewhere? I don't even need a UI I just need to be able to dump in a massive list of balanced sources.

PR #703, opened 2024-07-25
Enhance File Upload Component with Improved UI and Delete Functionality

muhammedsaidckr:
This pull request introduces several enhancements to the FileUpload component. Below are the detailed changes:

Uploaded Files List:
-     Styled the list of uploaded files for a cleaner, more organized presentation.
-     Used Tailwind CSS classes to ensure consistent design and spacing across the component.

'n' to &apos;n&apos; because typescript error 

Added preventDefault in the delete button onClick event to prevent the modal from closing.

Screenshot from 2024-07-25 17-19-25

### DIFF
--- a/backend/server.py
+++ b/backend/server.py
@@ -77,6 +77,7 @@ async def websocket_endpoint(websocket: WebSocket):
                 json_data = json.loads(data[6:])
                 task = json_data.get("task")
                 report_type = json_data.get("report_type")
+                source_urls = json_data.get("source_urls")
                 tone = json_data.get("tone")
                 headers = json_data.get("headers", {})
                 filename = f"task_{int(time.time())}_{task}"
@@ -86,7 +87,7 @@ async def websocket_endpoint(websocket: WebSocket):
                 report_source = json_data.get("report_source")
                 if task and report_type:
                     report = await manager.start_streaming(
-                        task, report_type, report_source, tone, websocket, headers
+                        task, report_type, report_source, source_urls, tone, websocket, headers
                     )
                     # Ensure report is a string
                     if not isinstance(report, str):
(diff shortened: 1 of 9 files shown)


muhammedsaidckr:
@ElishaKay 
Yeah you are right that would be better uploaded_files_count == 0? then Uploaded files will be hidden. I will apply that. 

The app seems to get stuck on the words: "✍️ Writing summary for research task:..."  

I experienced the same situation, but It responds even though it is late. I think there is a problem on the "openai" side right now. https://status.openai.com/

ElishaKay:
> @ElishaKay Yeah you are right that would be better uploaded_files_count == 0? then Uploaded files will be hidden. I will apply that.
> 
> The app seems to get stuck on the words: "✍️ Writing summary for research task:..."
> 
> I experienced the same situation, but It responds even though it is late. I think there is a problem on the "openai" side right now. https://status.openai.com/

You may be on to something there - same error on master

ElishaKay:
@FarVision2 made a good point here about [filtering by domain name here ](https://github.com/assafelovic/gpt-researcher/pull/700#issuecomment-2246245830)

How about: if the Selected Retriever supports filtering Search Results by domain, that option will be available via the UX as well?

For example, if the retriever=google, appending filters to the query should work:
`site:lifewire.com OR site:nasa.gov electric vehicles`

muhammedsaidckr:
Screenshot from 2024-07-26 16-31-59

Necessary changes have been applied to uploaded files @ElishaKay

muhammedsaidckr:
> @FarVision2 made a good point here about [filtering by domain name](https://github.com/assafelovic/gpt-researcher/pull/700#issuecomment-2246245830)
> 
> How about: if the Selected Retriever supports filtering Search Results by domain, that option will be available via the UX as well?
> 
> For example, if the retriever=google, appending filters to the query should work: `site:lifewire.com OR site:nasa.gov electric vehicles`

I'll look into this and see what we can do.

ElishaKay:
all tests pass

ElishaKay (approved):

Issue #684, opened 2024-07-18
TypeError: the JSON object must be str, bytes or bytearray, not coroutine


ruzicka02:
Greetings, I have encountered the following issue when running the GPT Researcher from the provided Python script `multi_agents/main.py`, inserting the part of stack-trace considered relevant:
```
  File ".../site-packages/multi_agents/agents/reviser.py", line 46, in run
    revision = self.revise_draft(draft_state)
  File ".../site-packages/multi_agents/agents/reviser.py", line 42, in revise_draft
    return json.loads(response)
  File ".../json/__init__.py", line 339, in loads
    raise TypeError(f'the JSON object must be str, bytes or bytearray, '
TypeError: the JSON object must be str, bytes or bytearray, not coroutine
 ```
This clearly leads to some issue regarding the use of asyncio, as `response` is returned from async function `call_model`. I tried a few different changes to fix that issue, but my understanding of asyncio use in this context is rather basic. Could someone please check this out?

assafelovic:
This seems related to the response from the reviser agent. What LLM provider are you using?

ruzicka02:
I am using OpenAI.

As far as I understand, the return value type of async function is:
1. coroutine, if the function is called without await,
2. the actual value, if called with await.

But any of my attempts to fixed this just caused problems elsewhere

emilmirzayev:
@ruzicka02 , some monkeypatch is to change `ReviserAgent` in `reviser.py`. I tried it today and it worked partly. Albeit no errors it might end in a loop and langgraph might throw an error.

```python
class ReviserAgent:
    def __init__(self, headers=None):
        self.headers = headers or {}

    async def revise_draft(self, draft_state: dict):
        """
        Review a draft article
        :param draft_state:
        :return:
        """
        review = draft_state.get("review")
        task = draft_state.get("task")
        draft_report = draft_state.get("draft")
        prompt = [{
            "role": "system",
            "content": "You are an expert writer. Your goal is to revise drafts based on reviewer notes."
        }, {
            "role": "user",
            "content": f"""Draft:\n{draft_report}" + "Reviewer's notes:\n{review}\n\n
You have been tasked by your reviewer with revising the following draft, which was written by a non-expert.
If you decide to follow the reviewer's notes, please write a new draft and make sure to address all of the points they raised.
Please keep all other aspects of the draft the same.
You MUST return nothing but a JSON in the following format:
{sample_revision_notes}
"""
        }]

[...]

Issue #548, opened 2024-05-29
Multi-Agents will have redundant content across different sections


DandinPower:
Hello, I have tried the multi-agent implementation found in the `multi-agents` folder. I also read the multi-agent blog first. I discovered that even though each subsection focuses on different topics, it is easy for the content to overlap and discuss the same concepts across different subsections. As a result, the final report often contains a lot of redundant content, which is not useful at all.

I initially tried to add guidelines like "Each subsection must not have redundant content across different subsections." However, since the reviewer and reviser agents only work on individual subsections, the guidelines can only be applied at the subsection level, not to the entire report. Consequently, the reviewers and revisers are unaware of what is happening in other subsections, making it difficult to resolve the redundancy problem.

I am considering a solution where we have a chief reviewer and reviser. After each subsection completes its research, the chief reviewer and reviser would evaluate the final research across all subsections, ensure there is no redundant content, and provide revision directions to each subsection group to restart their research. 

I think this kind of workflow will make the whole process more complex, increase token waste, and cause higher latency. However, I believe that if we can set global guidelines, such as "Each subsection must not have redundant content across different subsections," it can improve the final report's robustness and usefulness.

assafelovic:
Hey @DandinPower that's a great discovery and definitely a huge improvement to the experience. At the moment (mostly for cost purposes) indeed the report does not take into consideration previous sub topics. To enable this every iteration would need to see the entire report generated before that which would be very costly in terms LLMs calls. I guess it's a tradeoff but definitely something we can consider adding as an option

antoremin:
Hey, I played with the researcher and noticed same issue. IMO having the tradeoff is well worth it, since the alternative is to remove redundant sections by hand. Maintaining coherence across the entire report seems paramount for real-world applications and complex subjects. 
Maybe it could be an option so users could choose. Excited about possible architecture iterations in this direction and would love to help!

DandinPower:
Hello @antoremin, I have forked the repo and written a brief draft about the modification design. If you are interested, you are welcome to check the `multi_agent_v2` folder for details !

assafelovic:
Hey @DandinPower would love help with a PR for this! Currently working on a new front end UX/UI

DandinPower:
Hello @assafelovic, thank you for your invitation! I am willing to help with a PR after I finish the feature about the workflow I mentioned before.

assafelovic:
[...]

PR #705, opened 2024-07-29
Async fix

ruzicka02:
Issue mentioned in https://github.com/assafelovic/gpt-researcher/issues/684, and multiple other issues hidden behind it. Now, multi_agents flow can finish even with `"follow_guidelines": true`. Set that as the new default in example `task.json`, as I assume it was turned off because of this issue.


assafelovic (approved):
@ruzicka02 You're awesome, this is super important! Can you help and update the related opened issues? Thank you for the contribution!

ruzicka02:
@assafelovic closed the one related issue (created by me) mentioned, I would say https://github.com/assafelovic/gpt-researcher/issues/548 is about a slightly different problem

PR #713, opened 2024-07-31
Feature: Reducing redundancy problems in Detailed Reports

DandinPower:
## Motivation

In the current `DetailedReport`, it is easy to get redundant content across different subtopics. This is because only `existing_headers` are used to ask the LLM to prevent generating duplicate subsections. In #548, I mentioned this problem and had some discussions to plan modifications to fix it.

## High-level explanation of solution

In my solution, the whole process can be described in the following steps:

For each subtopic report generation:

1. Pass the `global_written_contents`, which are written by previous subtopic reports, into the current subtopic researcher.
2. Based on the conducted context, generate draft section titles for later retrieval.
3. Use the current subtopic task + draft section titles to retrieve relevant written content from `global_written_contents`.
4. Write new subtopics based on the conducted context and consider previous relevant sections to avoid writing redundant content.
5. Extract the new subtopic report and store it in `global_written_contents`.

### Extract and Store (Previous Step 5)

1. To extract the written content from the report, I added a new `extract_sections` function in `gpt_researcher/master/actions.py`. This function is used to extract the written content from subtopic report and put the result in a list of dictionaries (each dictionary represents one subsection title and its content).
2. To store all written content, I added a new `self.global_written_sections: list` in `DetailedReport` to store all written subsections accumulatively.

This modification is implemented in commit `1bed6e1`.

### Generate draft titles (previous step 2)

1. To generate draft titles based on research context, I added a new function called `generate_draft_titles_prompt` in `gpt_researcher/master/prompts.py`. This function will generate a prompt for llm to generate a list of headers based on the context. You can see the detailed prompt in this file.
2. Added a generation helper function in `gpt_researcher/master/actions.py` called `generate_draft_section_titles`. This function will use the previous prompt generation function to get the prompt, and it will call `cfg.fast_llm_model` to generate a list of draft section titles. Since this task is not difficult but may have a lot of context, the better LLM choice is using a fast LLM rather than a smart LLM.
3. Added a new member function `get_draft_section_titles` in `GPTResearcher`. This function will call the previous generation helper function and log to inform the user that it is currently generating draft section titles.

This modification is implemented in commits `3e2df16` and `ed2b7cf`.

### Retrieve the relevant content (previous step 3)

1. I referenced the `context_compressor` implementation to implement the new `WrittenContentCompressor`. This class can use a query to find relevant written content based on embedding and similarity.
2. Added a `get_similar_written_contents_by_draft_section_titles` member function to `GPTResearcher` and its helper function. The concept of this function is to allow `DetailedReport` to use current subtopics and draft section titles to retrieve relevant written sections. The flow of this function is to use current subtopic and all title as multiple queries, use `WrittenContentCompressor` to retrieve relevant content asynchronously and in parallel, gather the retrieval results, and remove redundant retrieved sections. There are 3 hyperparameters that can be set: `overall max result`, `each query max result`, and `similarity threshold`. I currently set the default `overall max result`=`10`, `each query max result`=`10`, and `similarity threshold`=`0.5`. In my observation, we need to set a higher similarity threshold to prevent irrelevant content from occupying the return result and causing actual relevant content to be discarded due to the overall max result limitation (this limitation is used to prevent sending too much previous written content to the LLM as context).
3. I also added logging to show the final retrieval results, allowing users to monitor whether it's retrieving content that may cause redundancy issues.

This modification is implemented in commit `134ea86`.

### Consider previous written context when writing report (previous step 4)

1. Modify `DetailedReport` to pass the `relevant_written_contents` into the `write_report` function. In `write_report`, add additional logic for `subtopic_report` to pass `relevant_written_contents` into the prompt generation function.
2. The major modification is in the prompt generation function, located in `gpt_researcher/master/prompts.py`. Additional instructions have been added to allow the LLM to consider the relevant written content to prevent writing duplicate content. For detailed modifications, refer to the file directly.

These modifications are implemented in commits `7667d0c` and `ba9ff0a`.

## Results and Conclusion

I tested several research topics I'm familiar with and found that the redundancy issue occurs less frequently.

In my experiments, the redundancy issue in the original version followed this pattern: two subtopics with similar directions led to similar contexts, resulting in the generation of similar subsections (with many identical subsections, each discussing similar topics). This problem is less prevalent in my implementation because it considers previously written content. In the current implementation, the quality for the previous redundancy scenario is improved: two subtopics with similar directions still lead to similar contexts, but content is generated on different aspects (though there's still a chance of duplication in a few subsections). In this situation, I ensured that the previous retrieval part correctly retrieved the relevant content, so the root cause is that the LLM cannot follow the instructions 100%.

I think there's a chance to tune my current prompt to better utilize the `relevant_written_contents` information to assist the LLM in generating a non-redundant report. Additionally, my experiments may have limitations since there are many aspects and different research topics.

I am willing and expect everyone to test this feature and provide feedback to help tune the prompt or better follow best practices.

### DIFF
--- a/backend/report_type/detailed_report/detailed_report.py
+++ b/backend/report_type/detailed_report/detailed_report.py
@@ -138,6 +138,9 @@ class DetailedReport:
         # Conduct research on the subtopic
         await subtopic_assistant.conduct_research()
 
+        draft_section_titles = await subtopic_assistant.get_draft_section_titles()
+        parse_draft_section_titles = extract_headers(draft_section_titles)
+
         # Here the headers gathered from previous subtopic reports are passed to the write report function
         # The LLM is later instructed to avoid generating any information relating to these headers as they have already been generated
         subtopic_report = await subtopic_assistant.write_report(self.existing_headers)
--- a/backend/report_type/detailed_report/detailed_report.py
+++ b/backend/report_type/detailed_report/detailed_report.py
@@ -138,8 +138,13 @@ class DetailedReport:
         # Conduct research on the subtopic
         await subtopic_assistant.conduct_research()
 
+        # Use research results to generate draft section titles
         draft_section_titles = await subtopic_assistant.get_draft_section_titles()
         parse_draft_section_titles = extract_headers(draft_section_titles)
+        parse_draft_section_titles_text = [header.get("text", "") for header in parse_draft_section_titles]
+
+        # Use the draft section titles to get previous relevant written contents
+        relevant_contents = await subtopic_assistant.get_similar_written_contents_by_draft_section_titles(current_subtopic_task, parse_draft_section_titles_text, self.global_written_sections)
 
         # Here the headers gathered from previous subtopic reports are passed to the write report function
         # The LLM is later instructed to avoid generating any information relating to these headers as they have already been generated
--- a/backend/report_type/detailed_report/detailed_report.py
+++ b/backend/report_type/detailed_report/detailed_report.py
@@ -148,7 +148,7 @@ class DetailedReport:
 
         # Here the headers gathered from previous subtopic reports are passed to the write report function
         # The LLM is later instructed to avoid generating any information relating to these headers as they have already been generated
-        subtopic_report = await subtopic_assistant.write_report(self.existing_headers)
+        subtopic_report = await subtopic_assistant.write_report(self.existing_headers, relevant_contents)
 
         # Update the global written sections list
         self.global_written_sections.extend(extract_sections(subtopic_report))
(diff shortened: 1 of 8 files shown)


assafelovic (approved):

assafelovic:
Great job @DandinPower this feature definitely improves overall redundancy and is greatly written. Added some minor adjustments but this definitely sets up the ground for further improvements in detailed reports. Kudos!

DandinPower:
Thanks for your review! Looking forward to future improvements!

LI-YONG-QI:
@DandinPower  Awesome

PR #726, opened 2024-08-05
Add Support for Multiple Retrievers in GPTR

0x11c11e:
This pull request enhances GPTResearcher by introducing support for selecting multiple retrievers. The following changes have been made:

- Updated the `.env.example` file to include configurations for both `RETRIEVER` and `RETRIEVERS`.

  1. If `RETRIEVER` is set and `RETRIEVERS` is not set, the value of `RETRIEVER` will be used by default.
  2. If `RETRIEVERS` is set, the values specified in `RETRIEVERS` will be used, separated by commas.
  3. If both are set, `RETRIEVERS` takes precedence.

- Refactored the search result handling to iterate through multiple retrievers and combine their results.

- Updated the configuration to support multiple retrievers by parsing the `RETRIEVERS` environment variable.

- Added helper functions to manage the retrieval process based on the new configuration.

These changes allow for greater flexibility and customization in selecting data sources, improving the overall capability and robustness of GPTResearcher.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -18,7 +18,10 @@ OLLAMA_EMBEDDING_MODEL=
 # The Ollama endpoint to use
 OLLAMA_BASE_URL=
 
-# Choose one of the available retrievers by uncommenting the desired retriever:
+# Choose one or more of the available retrievers by uncommenting the desired retriever(s).
+# If there is one RETRIEVER, it reads only that one. 
+# If there are multiple RETRIEVERS, separate them with commas (e.g., RETRIEVER=tavily,arxiv).
+# Available retrievers:
 # RETRIEVER=arxiv
 # RETRIEVER=bing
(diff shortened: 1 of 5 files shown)


assafelovic:
Hey @0x11c11e I think it's better to name it `CUSTOM` instead of `RETRIEVERS`. Probably a bit more explainable.

assafelovic:
Hey @0x11c11e I have an idea that might be more simple. How about instead of adding another env + specifying `RETRIEVERS` how about simply enabling adding comma separated retrievers inline? So basically the developer can either specify one or many retrievers. If we catch comma separated, we assume it's a list of retrievers and not a single retriever. Wdyt? Probably also make the code more readable and simple.

assafelovic (approved):
This is HUGE! Thank you @0x11c11e !

PR #738, opened 2024-08-08
moving the sources box above the logs that follow it

ElishaKay:
moving the sources box above the logs that follow it & white copy button

Tests ran & passed where report_type = summary or multi_agents


ElishaKay (commented):
Tests ran & passed where report_type = summary or multi_agents

assafelovic (approved):

PR #755, opened 2024-08-14
Stabilize master

ElishaKay:


ElishaKay:
Screen Shot 2024-08-14 at 20 21 25

hybrid option added to nextjs app & runs to completion - reading from both docs & web :)

assafelovic (approved):

PR #759, opened 2024-08-15
made use of azure_openai consistent (and not just in the base model, …

danieldekay:
The last update changed the string that identifies Azure OpenAI resources from `azureopenai` to `azure_openai`, but just for the base model, not for embeddings.

This change makes it consistent for both base model and embedding model.


assafelovic (approved):
Thank you super valuable!

PR #783, opened 2024-08-21
✅ ready for merge: Fullstack fixes

ElishaKay:
Tested human in the loop extensively for the multi_agents experience both when running via the CLI & NextJS Frontend.

This PR Supports:

handling all 3 potential inputs of HIL:
 - 'no', 
 - blank input
 - custom human feedback

The multi_agents can pivot dramatically based on the human_feedback 🎊

In future sprints, we'll want to:
 - create a function which will serve both the CLI & Frontend: `process_human_feedback`
 - enable testing the multi_agents flow via the relevant Github workflow


multi_agents/agents/editor.py
```diff
@@ -42,7 +40,6 @@ async def plan_research(self, research_state: dict):
             "role": "user",
             "content": f"Today's date is {datetime.now().strftime('%d/%m/%Y')}\n."
                        f"Research summary report: '{initial_research}'\n"
-                       f"{f'Human feedback: {human_feedback}. You must plan the sections based on the human feedback.' if include_human_feedback else ''}\n"
```
assafelovic, line 45: @ElishaKay why did you remove it? It's critical for updating the plan based on the human feedback

ElishaKay: There were two edge cases I needed to handle (when human_feedback = no or human_feedback = blank), therefore, I needed to refactor a bit

In other words, instead of processing the human_feedback in `editor_agent.plan_research` I processed it one function down in `editor_agent.run_parallel_research`

 Here's the gist of the new processing logic:
 
 ```
 human_feedback = research_state.get("human_feedback")

final_drafts = [
    chain.ainvoke({
        "task": research_state.get("task"),
        "topic": query + (f". Also: {human_feedback}" if human_feedback is not None else ""),
        "title": title,
        "headers": self.headers
    })
    for query in queries
]
```

It's a pretty solid approach, in my opinion, since the GPTR class will then factor in the human_feedback when generating the subqueries in the remainder of the flow

assafelovic (approved):

PR #784, opened 2024-08-21
Fix TypeError in custom reports due to missing 'tone' parameter

Gitzoz:
Resolved a `TypeError` that occurred when generating custom reports, caused by the `generate_custom_report_prompt()` function receiving an unexpected keyword argument 'tone'. The issue was due to the absence of the 'tone' parameter in the function definition, which has now been added to ensure correct functionality.


assafelovic (approved):
Thank you!

PR #786, opened 2024-08-21
Improved azure openai config

danieldekay:
I just ran into this problem: by default GPT-R assumes I have a GPT-4o-mini at hand, which we don't, because Microsoft doesn't offer it yet in Europe.

Digging a bit into the code I changed the following:

1. config.py -- allow for setting a default model
2. llms.md -- documented the env vars that need to be set
3. base.py -- let the .env file be loaded before the app is imported


assafelovic (approved):
Cheers @danieldekay !! This is great

PR #789, opened 2024-08-21
Enhancement: OpenAI Embedding Model Selection

eitag-uni:
**Description:** 
This pull request enhances the flexibility of the Memory module by enabling the use of various OpenAI embedding models.
* **What changes are being made?** 
Adding the ability to use different openAI embedding model 
Also calculating the estimate cost based on different embeddings models 
    

* **Why are these changes necessary?** 
    In the Memory module, embeddings.py 
    If the embedding provider is openAI the only value that is possible is  'text-embedding-3-small' 

* **How were these changes implemented?** 
    Added the ability to use OPENAI_EMBEDDING_MODEL as an environment variable


assafelovic (approved):
This is very cool and valuable thank you!

Issue #764, opened 2024-08-16
multi-agents: editor doesn't use LLM factory from GPT-researcher


danieldekay:
https://github.com/assafelovic/gpt-researcher/blob/14e7991b9be4b2456fcfc9490c3618ca67b1a009/multi_agents/agents/editor.py#L57

Instead, it calls a function in multi-agenst/utils/llms.py that is hard coded to use OpenAI models only.

I just tried to change that locally to something else, like:

`async def call_model(
    prompt: list,
    model: str,
    max_retries: int = 2,
    response_format: str = None,
) -> str:

    optional_params = {}
    if response_format == "json":
        optional_params = {"response_format": {"type": "json_object"}}

    lc_messages = convert_openai_messages(prompt)
    response = (
        AzureChatOpenAI(
            azure_deployment=model,
            model=model,
            max_retries=max_retries,
            model_kwargs=optional_params,
            api_key=SecretStr(os.getenv("AZURE_OPENAI_API_KEY")),
            azure_endpoint=os.getenv("AZURE_OPENAI_ENDPOINT"),
            openai_api_type="azure",
            temperature=0.6,
            max_tokens=4000,
        )
        .invoke(lc_messages)
        .content
    )
    return response`

But that only goes so far, as somewhere down the line it fails with:

`  File "/home/username/Envs/gpt-researcher-3.12.4/lib/python3.12/site-packages/langchain_openai/chat_models/base.py", line 629, in _generate
    response = self.root_client.beta.chat.completions.parse(**payload)
               ^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'beta'`

ElishaKay:
Good call @danieldekay 

Looks like we'll want to pass an ```llm``` argument into the call_model function

danieldekay:
That was my first guess, too. My hard-coded change, however, caused issues. My guess is actually that it's due to json output being required, and maybe I didn't call it correctly.

However, in some way, it doesn't make sense to not use the LLM factory that gpt-researcher core uses, right? I don't know yet enough about the code base, to make any suggestion as to where it makes the most sense. I guess the other agents all essentially do the basic gpt-research tasks, whereas the editor has a different setup...

danieldekay:
Update: I went down the rabbit hole, and realized that the documentation of langchain is not working, either.

[...]

PR #795, opened 2024-08-22
multi-agent approach should now use .env file for config of llm

danieldekay:
This addresses issue #764 

Changed the `call_llm`  function 
- to re-use the "select the right model from .env" from GPT-R
- Changed the handling of the response in "json" mode to return a parsed json, else return the response as is.
- Simplified the function to take less params, as most can/should be configured either via task or .env (Config() from GPTR, or config_file)

@ElishaKay, I tried to touch as little code as possible, but also don't fully grasp the langgraph logic you built. Please check If I broke something essential.

I can run and generate reports using multi-agent with AzureOpenAI now. (but my report switched from English to Spanish in the last agent -- my colleagues tell me GPT-4o likes to speak Spanish more than English and often does that.)

### DIFF
--- a/multi_agents/agents/editor.py
+++ b/multi_agents/agents/editor.py
@@ -6,10 +6,7 @@ import asyncio
 import json
 
 from ..memory.draft import DraftState
-from . import \
-    ResearchAgent, \
-    ReviewerAgent, \
-    ReviserAgent
+from . import ResearchAgent, ReviewerAgent, ReviserAgent
 
 
 class EditorAgent:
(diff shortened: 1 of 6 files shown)


ElishaKay:
@danieldekay that's awesome!
Great initiative, as usual :)

@assafelovic wrote the multi_agents backend so he's probably you're point of contact here

From my end, I can offer you an automated test that we can trigger before merging new PR's to master - in that way, if a pull request introduces a bug into running multi_agents with azure, our automated test should catch it

What I'll need to do that:
- your multi_agents/task.json
- an example env

Regarding localization: Sounds like you should add to the prompt: "I don't speak Spanish - please answer in English" - we'll need to search the codebase for existing localization features if we plan to extend / enforce the existing localization.

May the force be with you

danieldekay:
Haha, yes, I found my Spanish problem: it's inside `task.json` in the guidelines.

https://github.com/assafelovic/gpt-researcher/blob/c897a1f30df43f82df29918c97cd64107e1630cf/multi_agents/task.json#L15

assafelovic (approved):
@danieldekay this is awesome and will push us towards one synced system. 👏👏

PR #797, opened 2024-08-23
Update prompts.py

Einengutenmorgen:
Missing parameter: Tone in generate_outline_report,  Related to issue 791
Tested with todays Version


assafelovic (approved):
Great catch thank you!

PR #828, opened 2024-09-06
Adding Custom Configuration Setup Instructions to GPTR

monolok:
I’ve added a section at the end of the `gpt_researcher/README.md` to address frequent questions about setting up custom configurations. Noticing that many users had queries on this topic, I included detailed instructions for creating and using a custom JSON configuration file.

### DIFF
--- a/gpt_researcher/README.md
+++ b/gpt_researcher/README.md
@@ -45,4 +45,26 @@ if __name__ == "__main__":
 
 ```
 
+### Customize the configuration (optional)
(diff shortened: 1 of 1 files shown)


assafelovic:
Hey @monolok thanks! Can you also add the option of using env vars? If we're planning to add this we may as well provide the full overview.

monolok:
hey @assafelovic ☺️ Absolutely ! on it .

assafelovic (approved):
Thanks this is great!

PR #831, opened 2024-09-08
Fix Last Paragraph is not Sent to Websocket

khoapip:
How to reproduct: Ask a random question and the result shown in the webUI does not contain the last few lines when compare to the PDF.

Description: when the model output chunks of messages, the chunks are combined to paragraphs and send to websocket. However, the last paragraph is never sent (if the last paragraph does not contain \n). My change check if paragraph is not empty to send the last paragraph out.


assafelovic (approved):
Thanks and good catch!

PR #836, opened 2024-09-12
added an example on a custom configuration e.g. Azure

danieldekay:
**Enhancements:**

- Added a more concrete example of a non-OpenAI model configuration.
- Included a link to LangChain for further information on model parameters, as each LLM provider requires different environment keys.


assafelovic (approved):
Thanks as always @danieldekay this will definitely help users on onboard more easy

PR #837, opened 2024-09-12
Update server.py

danieldekay:
Just changed one line that was making problems in my setup with missing folder/existing folders.


assafelovic (approved):
Good catch!

PR #819, opened 2024-09-02
AI Dev Team

ElishaKay:
*Setup*: 

Step 1: Generate a Github Personal Access Token

```
For the GITHUB_TOKEN (Personal Access Token) value, you'll need to generate a Personal Access Token:
Click on your profile picture in the top-right corner of GitHub and select "Settings".
In the left sidebar, click on "Developer settings".
Click on "Personal access tokens", then "Tokens (classic)".
Click "Generate new token" and select the appropriate scopes for your needs.
Copy the generated token.
Paste the token you generated into the value field of the new secret.
Click "Add secret" to save it.
```

Step 2: Set these environment variables:
```bash
OPENAI_API_KEY={Your OpenAI API Key here}
TAVILY_API_KEY={Your Tavily API Key here}

PGVECTOR_CONNECTION_STRING=postgresql://username:password...
GITHUB_TOKEN=
```

Step 3:

`pip install -r multi_agents/requirements.txt`
`python -m multi_agents.dev_team.main`

*Status*: Running to completion

PR #838, opened 2024-09-12
Add Document To Vector Store

khoangothe:
Documents, crawled urls, and website will be chunked and loaded to the inputted vector store if vector_store is not None. Although adding the data  in Document would be more efficient, but I think this solution makes the code decoupled and easier to maintain. 

By default this changes won't add any features, but new applications can be implemented based on the vectorstore (like chatting with the sources)


khoangothe:
Would be cool to have this idea reviewed! I'll add a script test if I am allowed to proceed

assafelovic:
Hey @khoangothe this is a great direction! Can you share an example of how it can be used?

khoangothe:
Hi @assafelovic, Thanks for the review! I just added a commit to document how it should be used. Here's the code that I used to test locally. 

`from gpt_researcher import GPTResearcher

from langchain_community.vectorstores import InMemoryVectorStore
from langchain_openai import OpenAIEmbeddings

vector_store = InMemoryVectorStore(embedding=OpenAIEmbeddings())

query = "The best LLM"

# Create an instance of GPTResearcher
researcher = GPTResearcher(
    query=query,
    report_type="research_report",
    report_source="web",
    vector_store=vector_store, 
)

# Conduct research, the context will be chunked and store in the vector_store
await researcher.conduct_research()

# Query the 5 most relevant context in our vector store
related_contexts = await vector_store.asimilarity_search("GPT-4", k = 5) 
print(related_contexts)
print(len(related_contexts)) #Should be 5 `

assafelovic:
Thanks @khoangothe excuse me if I might be missing something but how is this different than this? https://docs.gptr.dev/docs/gpt-researcher/context/vector-stores

khoangothe:
@assafelovic  Sorry if my examples were not clear enough. The one you link allows you to  talk to your vector store, so your vector store needs to already have information for gpt-researcher to do research on (when `report_source` is set to `langchain_vectorstore`). My changes will allow GPT-Researcher to add new reports to the vector store, so the scraped website + documents will be added to your vector store, so later you can reuse your vector store for other purpose, like RAG.

In the example, starting with an empty `InMemoryVectorStore` right after `await researcher.conduct_research()`, the vector store will have everything stored. and `related_context` now contains information that was scraped

hslee16:
Looks good 👍🏼

ElishaKay:
@assafelovic 

This idea of persistent & re-usable vector storage that can be leveraged across reports & follow-up questions is very interesting for me.

I've merged this branch into #819 & am planning on testing extensively with PGVector Storage.

khoangothe:
@ElishaKay  @assafelovic  Hi guys, I was able to resolve the merge conflict and provided tests cases for the scenarios that I implemented. For each type of knowledge source (urls, hybrid, local, webs, langchain documents) data will be ingested in the vector_store that user provided, usage is written in the tests. I also raised 1 issue in Discord. 
It would be cool to have this pr reviewed, tested and hopefully merged! I want to also implement chatting with Data Source but it currently depends on this PR.  Thanks guys for your help!

assafelovic (approved):
This is awesome @khoangothe kudos for the hard work and implementation. Looking forward to the next PRs that can empower this

danieldekay:
Love the idea, as you can build out a knowledge base through various queries this way. It adds a bit more human-in-the-loop for complex topics.

One use case could be if you already have a corpus of literature, but want to add more recent content via GPT-R's searches.

PR #841, opened 2024-09-13
Fix SyntaxError in prompts.py due to conflicting quotes in f-string

onlydole:
## Description
This PR fixes a SyntaxError in `/usr/src/app/gpt_researcher/master/prompts.py` that was causing the application to fail on startup. The error was due to conflicting quotes in an f-string used to format the current date.

## Changes Made
- Modified line 33 in `/usr/src/app/gpt_researcher/master/prompts.py` to resolve quote conflicts in the f-string.


assafelovic (approved):
Thank you good catch!

PR #843, opened 2024-09-15
Simple Docker files optimization.

yigit353:
The `version` tag at the beginning of the `docker-compose.yml` file is obsolete, so I just removed it.

Also, in `Dockerfile`, `FROM` and `as` should have the same capitalization, so I capitalized `as` into `AS` for consistency.


ElishaKay (approved):
pulled & tested.
works as expected

PR #832, opened 2024-09-09
[Experimental] Chat with history

khoangothe:
I am trying to implement chat history to chat with the final report. The code is just experimental since I'm not sure whether this is where I should put the chat agent. The code will put the report and vectorize in a inmemory vectorstore to get the relevant chunks and I used default LangGraph React Agent. Would be cool if I could have some feedbacks on how to proceed! Thanks guys

Here's the UI (default to chatting after having a report, pressing the search symbol will conduct another research)
image

PR #885, opened 2024-10-05
Chat with History

khoangothe:
Continue on #832 but merge with the newer release of gpt-researcher. I think the plan for this branch would be to allow user to chat with the report. When we have a vector_store, we can chat with the crawl data too.


ElishaKay (approved):

ElishaKay:
Tested and works great

PR #889, opened 2024-10-07
Use apiVariables.LANGCHAIN_API_KEY instead of hardcoded authToken in Langgraph.js

jcalebsmith:
Removed the hardcoded token in Langgraph.js and replaced it with LANGCHAIN_API_KEY variable from the modal screen.


ElishaKay:
Hey @jcalebsmith 

The value to pass in that request header is the langgraph_cloud auth token which you can get via the relevant deployment's Dashboard view

I'll share a screenshot when I get a chance 

Meanwhile, I'm checking with the Langchain team on this multiple subgraph limitation

Will update here when there's news

PR #928, opened 2024-10-19
Update vector-store.py

DhanushNehru:


assafelovic (approved):
Hey @DhanushNehru check out the latest release. Refactored a massive part of the codebase. Shouldn't effect this feature but FYI for future updates

PR #931, opened 2024-10-20
update langchain huggingface embedding

hereiamravi:
this is causing the no content generation when google/huggingace are used as llm/embedding models
image


assafelovic (approved):

assafelovic:
Thanks super important!

PR #918, opened 2024-10-17
Specify embedding provider and model

kesamet:
- Allows user to specify embedding provider and model in the form `EMBEDDING="<embedding_provider>:<embedding_model>"`. Default is set `EMBEDDING="openai:text-embedding-3-large"`. This is similar to `FAST_LLM` and `SMART_LLM`
- Make deprecation warnings more intuitive
- Allows user to specify additional kwargs such as `llm_kwargs` and `embedding_kwargs` in an user-specified json file in `config_path`
- Clean up `gpt_researcher/config/config.py`. Eg `self.load_config_file` is removed as it is no longer in use

PR #933, opened 2024-10-20
"hack" to fix AzureOpenAI configuration problem

danieldekay:
Explanation:

Unfortunately, while all other LLM providers require only MODELNAME and APIKEY, Azure requires also a MODELDEPLOYMENT.
Since each model has its own deployment, when we use smart_model and fast_model, we need deployments for each.

The fix works, but only if the model name and deployment name are identical.

TODO: I did not test the embeddings model yet, as I cannot test it due to a throttled Azure Subscription at home (free trial tier)


assafelovic (approved):

danieldekay:
and maybe https://github.com/assafelovic/gpt-researcher/pull/918 fixes the Embeddings model to work with Azure, as well.

danieldekay:
confirmed: #918 fixes the embeddings with Azure.

PR #941, opened 2024-10-23
Feature/strategic llm

assafelovic:


assafelovic:
GPT Researcher just got smarter 🧠! It now leverages GPT o1 to make a better planning strategy for the research task. We're introducing a new layer of reasoning using the env var `STRATEGIC_LLM` which defaults to `openai:o1-preview` but can be changed to any of your preferred provider and model such as Anthropic's latest claude 3.5 sonnet. Pull the latest to enable this feature.

PR #944, opened 2024-10-23
Fix/bing retriever consistency

ibuder:
This fixes an inconsistency in the return type of the Bing retriever.

It could return a list or `None`. However, the [calling code](https://github.com/assafelovic/gpt-researcher/blob/1a36430acfc74d1804d313d20c958140a05f953d/gpt_researcher/skills/researcher.py#L284) assumed the result was always a list. As a result, the entire researcher would fail if the Bing retriever reached any code path that returned `None`.

This PR makes the Bing retriever always return a list, consistent with other retrievers.


assafelovic (approved):
Thanks super helpful!

PR #953, opened 2024-10-26
Sort pubmed_central search results by relevance versus default newest

cannin:
By default, it looks like pubmed_central results are sorted by latest. This can produce results that are not helpful. 

The added sort=relevance parameter can be tried with a link like this: 

https://eutils.ncbi.nlm.nih.gov/entrez/eutils/esearch.fcgi?db=pmc&term=what+are+the+physiological+functions+of+calcium+AND+free+fulltext%5Bfilter%5D&retmax=3&usehistory=y&retmode=json&sort=relevance


assafelovic (approved):
Thank you great improvement!

PR #958, opened 2024-10-27
Documentation update to correct LLM provider prefix for Azure OpenAI.

scchengaiah:
Updated LLM provider prefix from `openai` to `azure_openai`.


assafelovic (approved):
Thanks for resolving this, huge pain for some developers 🙏

PR #959, opened 2024-10-28
Change Prompt to Avoid Incorrect Information for PubMed Central Sources

cannin:
Currently, the LLM context does not have author or journal information available for PMC sources. The LLM tries to accommodate the user using names it finds in the text (e.g., authors of references cited by the source). This causes mismatches such as article authors and cited PMC sources in reports. The commit allows the LLM to provide a simpler but often more correct citation with just title and link.

### DIFF
--- a/gpt_researcher/prompts.py
+++ b/gpt_researcher/prompts.py
@@ -72,6 +72,31 @@ Every url should be hyperlinked: [url website](url)
 Additionally, you MUST include hyperlinks to the relevant URLs wherever they are referenced in the report: 
 
 eg: Author, A. A. (Year, Month Date). Title of web page. Website Name. [url website](url)
+
+# SPECIAL INSTRUCTIONS FOR PMC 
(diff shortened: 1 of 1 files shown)


assafelovic (changes requested):
@cannin since this is only relevant to pubmed retrievers I'd like to add a rule to only add this prompt when retriever = pubmed. I can get to it later in the week to accommodate this PR.

assafelovic:
Thanks for this contribution, and apologies for the long delay in getting to it. 🙏

As part of a backlog cleanup we're closing PRs that have been open for **6+ months** and have drifted substantially from the current `main` branch, which has since had major refactors (cost tracking, deep-research pipeline, retriever changes).

This is **not** a rejection of the idea. If it's still relevant, please rebase onto the latest `main` and reopen (or open a fresh PR) — we'll be much quicker this time. Thank you for your patience and for contributing to GPT Researcher!

Issue #909, opened 2024-10-14
SearXNG: Exception: Tavily API key not found. Please set the TAVILY_API_KEY environment variable.


PieBru:
**Describe the bug**
This is the log:
```log
INFO:     [11:03:05] 
🔍 Running research for ...
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/home/admin/Github/gpt-researcher/gpt_researcher/retrievers/tavily/tavily_search.py", line 39, in get_api_key
    api_key = os.environ["TAVILY_API_KEY"]
              ~~~~~~~~~~^^^^^^^^^^^^^^^^^^
  File "<frozen os>", line 679, in __getitem__
KeyError: 'TAVILY_API_KEY'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/admin/Github/gpt-researcher/venv/lib/python3.11/site-packages/uvicorn/protocols/websockets/websockets_impl.py", line 242, in run_asgi
    result = await self.app(self.scope, self.asgi_receive, self.asgi_send)  # type: ignore[func-returns-value]
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/admin/Github/gpt-researcher/venv/lib/python3.11/site-packages/uvicorn/middleware/proxy_headers.py", line 60, in __call__
    return await self.app(scope, receive, send)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/admin/Github/gpt-researcher/venv/lib/python3.11/site-packages/fastapi/applications.py", line 1054, in __call__
    await super().__call__(scope, receive, send)
  File "/home/admin/Github/gpt-researcher/venv/lib/python3.11/site-packages/starlette/applications.py", line 113, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/home/admin/Github/gpt-researcher/venv/lib/python3.11/site-packages/starlette/middleware/errors.py", line 152, in __call__
    await self.app(scope, receive, send)
  File "/home/admin/Github/gpt-researcher/venv/lib/python3.11/site-packages/starlette/middleware/cors.py", line 77, in __call__
    await self.app(scope, receive, send)
  File "/home/admin/Github/gpt-researcher/venv/lib/python3.11/site-packages/starlette/middleware/exceptions.py", line 62, in __call__
[...]

PR #960, opened 2024-10-28
Searx retriever refactoring

igochkov:
In relation to #909 

Searx retriever (searx.py) has been rewritten to conform to the SearXNG API specification. Searx is being discontinued and SearXNG is the successor.

N.B. To work with this retriever, the SearXNG service must be configured to enable the `json` output format. By default, for self-hosting scenarios, only `html` format is enabled in the settings.yml file!


assafelovic (approved):
This is great thank you!

PR #967, opened 2024-10-31
Add a history page (static)

niuguy:
The idea is simply to add a static history page and store the generated reports in a local SQLite database. Here are some screenshots for reference:

US election

US Election 2024 A Comprehensive

This is more of a proposal, but I think it might be more straightforward to submit a PR. Looking forward to the feedback!

### DIFF
--- a/backend/server/database.py
+++ b/backend/server/database.py
@@ -1,8 +1,10 @@
 import sqlite3
 from datetime import datetime
 
+DB_NAME = 'research_history.db'
+
 def init_db():
-    conn = sqlite3.connect('research_history.db')
+    conn = sqlite3.connect(DB_NAME)
     c = conn.cursor()
     c.execute('''
         CREATE TABLE IF NOT EXISTS research_history (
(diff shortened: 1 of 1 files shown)


ElishaKay:
Love it.

I'm voting for Postgres as the default data store, since it's also supported by [langchain vector stores - see PGVector](https://python.langchain.com/v0.2/docs/integrations/vectorstores/)

But, great proof of concept :)

niuguy:
@ElishaKay Yes, integrating with Postgres is certainly viable if that’s the plan. I’m using SQLite for now just for simplicity—it’s just a single file for the database, after all

muditjuneja:
Any update on this PR?

backend/server/database.py
```diff
@@ -0,0 +1,39 @@
+import sqlite3
+from datetime import datetime
+
+def init_db():
+    conn = sqlite3.connect('research_history.db')
```
muditjuneja, line 5: Can you move the db name as a const at the top and reuse the same name in other functions?

niuguy: sure

muditjuneja (approved):
Looks good for starters. 
Maintainers wdyt?

Issue #936, opened 2024-10-21
Azure Embedding Quota Limit


danieldekay:
**Describe the bug**
I am running a **detailed report** with Azure Openai, and am hitting quota limits. While I have a rate limit activated of 500k tokens per minute, it seems to still throw an error and not handle the throttling request well.

`openai.RateLimitError: Error code: 429 - {'error': {'code': '429', 'message': 'Requests to the Embeddings_Create Operation under Azure OpenAI API version 2024-02-15-preview have exceeded call rate limit of your current OpenAI S0 pricing tier. Please retry after 86400 seconds. Please go here: https://aka.ms/oai/quotaincrease if you would like to further increase the default rate limit.'}}
`

Of course the error message should be false, as my rate limits are not per day, and waiting 24h is not an option.

**Expected behavior**
- The log to the user (e.g. websocket) should indicate that there is throttling in place.
- The embedding should resume after a while, or indicate what the user can do instead.

roninio:
i have the same issue.

ElishaKay:
Fair point, we'll have to think how to make this smoother.

A) @danieldekay, is it the same docs that you're running reports on.

Have a look at this PR: "Documents, crawled urls, and website will be chunked and loaded to the inputted vector store if vector_store is not None."

https://github.com/assafelovic/gpt-researcher/pull/838

Meaning, if you run GPTR with the same Langchain vectorstore, perhaps it will cut down the embeddings processes.

B) the "cooling off" feature is also a good idea. Did you mention somewhere that there's a Langchain method we can leverage to get the required "cool off" period?

Once we have that, we can go about adding the websocket message. Adding an exception handle would also be a good first step which publishes a websocket message to the frontend

ElishaKay:
Awesome.

Adding to the [resilience channel on Discord](https://discord.com/channels/1127851779011391548/1301452566118400030).

For anyone reading who hasn't joined the Discord, [Join here to access the above link](https://discord.com/invite/QgZXvJAccX)

roninio:
I solved the issue by 
 _embeddings = AzureOpenAIEmbeddings(
                    model=model,
                    timeout=60,
                    chunk_size=1000,
                    azure_endpoint=os.environ["AZURE_OPENAI_ENDPOINT"],
                    openai_api_key=os.environ["AZURE_OPENAI_API_KEY"],
                    openai_api_version=os.environ["AZURE_OPENAI_API_VERSION"],
                    **embdding_kwargs,

[...]

PR #979, opened 2024-11-11
Azure embedding quota limit

roninio:
Fix the issue by explicitly setting the embedding model to "text-embedding-3-large" when using Azure openai


ElishaKay (approved):
[Based on the issue](https://github.com/assafelovic/gpt-researcher/issues/936)

PR #1007, opened 2024-12-06
Massive refactor to enable "black box" retrievers

assafelovic:


assafelovic:
Triage note — this one is **not** superseded, which makes it more interesting than #1242.

Opened 2024-12. The abstraction it proposes never landed: there's still no `gpt_researcher/retrievers/base.py` on `main`, and the retriever count has grown from a handful to **23**, with six more proposed in open PRs (#1814, #1850, #1867, #2015, #2048, #2068).

So the problem it was solving got considerably worse while it sat. #2074 also landed roughly twenty separate fixes for retrievers mishandling their own provider payloads — largely the same defensive-parsing bug repeated per retriever, which is exactly what a shared base class would centralise.

The diff itself is stale, but the *design question* is the one blocking six vendor-retriever PRs right now. A black-box retriever interface plus a documented plugin path would turn each of those from a bespoke merge decision into someone else's package. Worth reviving as a design rather than a rebase.

assafelovic:
Went back through this properly rather than treating it as stale, and it turns out to be the opposite of stale — the design is more relevant now than when you wrote it.

**What it proposes:** a `BaseRetriever` ABC in `gpt_researcher/retrievers/base.py` with an explicit contract, and critically a `requires_scraping` property — a retriever declares whether it returns *URLs that must be scraped* or *content it already fetched*.

**What `main` does instead, today:** guesses. `skills/researcher.py:855`:

```python
if url and raw_content and len(raw_content) > 100:
```

A 100-character threshold decides whether a retriever already fetched the full page.

**That guess became a bug.** #1846 and #1892 both reported search snippets being treated as pre-fetched full content, so URLs were never scraped and reports shipped with zero verifiable citations. Fixed in #2074 by capping snippet length inside the Searx and DuckDuckGo retrievers — the code comment there explains it is working around this exact threshold. So the fix is per-retriever tuning to game a heuristic that `requires_scraping` would have made a declaration.

Two more things point the same way:
- There is still no `retrievers/base.py` on `main`, and the retriever count has gone from a handful to **23**, with six more proposed in open PRs (#1814, #1850, #1867, #2015, #2048, #2068).
- #2074 landed roughly **twenty** separate fixes for retrievers mishandling their own provider payloads — largely the same defensive-parsing bug written 20 times, which is what a shared base class centralises.

The diff itself is 20 months old and won't apply. But the design question it raises is currently blocking six vendor-retriever PRs, and a documented base class plus a plugin/entry-point path would turn each of those from a bespoke merge decision into someone else's package.

Worth reviving as a design, not a rebase.

assafelovic:
Follow-up: the core idea from this PR has landed as #2079 (merged, `6f998577`).

`BaseRetriever.requires_scraping` is now in `gpt_researcher/retrievers/base.py`, and `_search_relevant_source_urls()` honours the declaration instead of guessing from `len(raw_content) > 100`. Five retrievers declare it so far — Tavily, Searx and DuckDuckGo as `True`, PubMed Central and Custom as `False`.

What was **not** taken from here, deliberately: the key-name normalisation and forced inheritance. Undeclared retrievers keep the exact legacy behaviour, so third-party and user-defined retrievers are untouched — that kept the change additive rather than breaking, which matters at this project's install base.

Closing this one as superseded in substance. The remaining 18 shipped retrievers can be annotated incrementally, and the abstraction is now there for new ones to build against — which was your point 20 months ago.

PR #1009, opened 2024-12-07
Fix max_results for Google Search

winsonluk:
There's a bug in Google Search where the max_results value in configs aren't respected. Found the bug and fixed it.


winsonluk:
@assafelovic just an easy one line bug fix for when you get a chance to review, thanks

assafelovic (approved):
Thank you!

PR #1013, opened 2024-12-09
Add support for DashScope provider in LLM and embeddings modules

AStupidBear:


assafelovic (approved):
This is great, thank you! How is this different from existing embeddings?

PR #1019, opened 2024-12-14
Default to the retriever set in config rather than just tavily

winsonluk:
```
-        retriever_env = os.environ.get("RETRIEVER", "tavily")
+        retriever_env = os.environ.get("RETRIEVER", config.get("RETRIEVER", "tavily"))
```

Right now if the retriever isn't set in the headers, it defaults to the environment variable RETRIEVER. If that's not set, it defaults to Tavily.

The problem with this is even if the retriever is set in the config, it's completely ignored.

I had no RETRIEVER environment variable set, but I had retriever set in my configs to Bing, Google, etc.,  and was wondering why I was blowing past all my Tavily API limits...

The current workaround is to manually set the retriever in the headers rather than the config.

### DIFF
--- a/gpt_researcher/config/config.py
+++ b/gpt_researcher/config/config.py
@@ -33,7 +33,7 @@ class Config:
             setattr(self, key.lower(), value)
 
         # Handle RETRIEVER with default value
-        retriever_env = os.environ.get("RETRIEVER", "tavily")
+        retriever_env = os.environ.get("RETRIEVER", config.get("RETRIEVER", "tavily"))
         try:
             self.retrievers = self.parse_retrievers(retriever_env)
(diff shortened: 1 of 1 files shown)


winsonluk:
@assafelovic another one :)

would appreciate a deployment of a new version when you get a chance, cheers!

assafelovic (approved):
Great point, thank you!

assafelovic:
@winsonluk just release a new version both in github repo and PIP package. Thank you for all your efforts, I honestly would love and appreciate additional PRs, the more the better! This is an open source project dedicated to democratizing agentic research, so it relies on improvements like yours. Thank you!

PR #1036, opened 2024-12-22
Prints unexpected Google retriever search API response

ivarprudnikov:
Google retriever silently fails to pull the links from the search API with error responses, i.e. 429. It took quite some time to understand where the failure happens because the UI just kept generating content without actually finding anything using the API. This PR to makes sure unexpected response gets logged thus making it easier to debug the running server.


assafelovic (approved):
Thank you!

PR #1044, opened 2024-12-26
Adding Tavily extract

pulvedu:
Adding Tavily extract as an available scraper. The scraper was added as an additional folder in `gpt_researcher/scraper/`. Tavily extract requires the tavily-python package which is not currently in the requirements.txt, ` _check_pkg()` function was added to `gpt_researcher/scraper/scraper.py ` to dynamically install the package if the scraper is invoked.


assafelovic (approved):
This is great! Will boost production usage for GPTR!

PR #1045, opened 2024-12-27
Support Bedrock embedding model

moritalous:
Add support for the Bedrock embedding model, expanding the list of supported embeddings and integrating the necessary imports and initialization logic.


assafelovic (approved):
Thank you this is a great addition!

PR #1048, opened 2024-12-28
Fix Gemini bug when system role is provided

winsonluk:
Gemini doesn't support role=system.

When `"SMART_LLM": "google_vertexai:gemini-1.5-flash-002"` for example, the current report generation fails with "Error in generate_report".

This fix adds a fallback. It first attempts the original chat completion:
```
            messages=[                                                                              
                {"role": "system", "content": f"{agent_role_prompt}"},                              
                {"role": "user", "content": content},                                               
            ],
```

Upon failure, it retries with this new chat completion:
```
                messages=[                                                                          
                    {"role": "user", "content": f"{agent_role_prompt}\n\n{content}"},               
                ],
```

This results in a successful report generation when `"SMART_LLM": "google_vertexai:gemini-1.5-flash-002"`.

### DIFF
--- a/gpt_researcher/actions/report_generation.py
+++ b/gpt_researcher/actions/report_generation.py
@@ -245,7 +245,7 @@ async def generate_report(
             llm_kwargs=cfg.llm_kwargs,
             cost_callback=cost_callback,
         )
-    except Exception as e:
+    except:
         try:
             report = await create_chat_completion(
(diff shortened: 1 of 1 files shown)


winsonluk:
@assafelovic short fix for when SMART_LLM fails with Gemini, thanks

assafelovic (approved):
Thanks for this!

PR #1056, opened 2024-12-31
Add DeepSeek to SUPPORTED_PROVIDERS

winsonluk:
This change adds `deepseek:deepseek-chat`, which is the latest DeepSeek-V3 model (released December 26, 2024).

DeepSeek doesn't have its own LangChain, so official DeepSeek documentation suggests using OpenAI's LangChain (https://api-docs.deepseek.com/faq), which this change implements.

Confirmed working with `FAST_LLM="deepseek:deepseek-chat"`, `SMART_LLM="deepseek:deepseek-chat"`, and `STRATEGIC_LLM="deepseek:deepseek-chat"` when `DEEPSEEK_API_KEY` system variable is set.

### DIFF
--- a/backend/server/server.py
+++ b/backend/server/server.py
@@ -41,6 +41,7 @@ class ConfigRequest(BaseModel):
     SERPER_API_KEY: str = ''
     SEARX_URL: str = ''
     XAI_API_KEY: str
+    DEEPSEEK_API_KEY: str
 
 
 # App initialization
--- a/docs/docs/gpt-researcher/llms/llms.md
+++ b/docs/docs/gpt-researcher/llms/llms.md
@@ -251,6 +251,15 @@ STRATEGIC_LLM="xai:grok-beta"
 ```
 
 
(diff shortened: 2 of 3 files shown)


winsonluk:
@assafelovic this adds DeepSeek as an additional provider since their DeepSeek-V3 benchmarks are pretty impressive

assafelovic (approved):
Love this! How's the model performance?

winsonluk:
> Love this! How's the model performance?

@assafelovic so far so good! I'm integrating it into https://ithy.com

It's a perfect candidate for `FAST_LLM` (faster and cheaper than 4o-mini / 3.5 Haiku / 1.5 Flash). When used as `SMART_LLM` or `STRATEGIC_LLM`, performance even exceeds most frontier models (though it starts to underperform when context tokens get very large).

My only concern is that they don't have published rate limits, but I haven't gotten any 429s so far 🤞

PR #1058, opened 2025-01-03
feature(configs): enhance temperature management via configuration (i.e. init for a new configuration system)

kga245:
Author note:
- These changes are a start. 
- They are not comprehensive. 
- I have tested to make sure that vanilla installs are in tact and that the few prescribed configuration changes made via this new method work (i.e. specifying LLMs and their temperature changes). 
- The main innovation here is that configuration now happens in a configs directory. You invoke the configuration by pointing to the one you like in you .env file. Keep as many configs as you like in your configs folder and freely swap between them. Use your .env file in a ci/cd pipeline to choose which configs drive which environments. 
- The old way was to use the gpt_researcher/config/variables/test_local.json and gpt_researcher/config/variables/default.py files to run your configs. I felt like this was core functionality that we as maintainers of the core would like to retain. End developers would need something more guardrails and would lend itself to better exploration. 
- My next steps are to methodologically go through all of the configs and make sure they are modifiable without breaking anything!
- Which brings up test. I added a slue of new tests. Use them if you want to test that your configs are working. 
- Documentation? Yeah not yet. Sorry. My brain is hurting. 
- Logging? There's some new stuff in there for sure. And it works. But I haven't formalized it into the new logging system yet. Look for it in the console primarily now and in the various tests that were built. 

Core Changes:
- Add granular temperature controls for different LLM types
- Implement temperature validation (0-1 range)
- Add hierarchical temperature inheritance
- Create sample_config.json as reference configuration

Configuration Updates:
- Add FAST_LLM_TEMPERATURE, SMART_LLM_TEMPERATURE, STRATEGIC_LLM_TEMPERATURE
- Update BaseConfig with new temperature settings
- Improve environment variable handling
- Add validation for temperature values

Test Coverage:
- Add comprehensive temperature configuration tests
- Test temperature inheritance and defaults
- Test invalid temperature handling
- Add query-based temperature selection tests

Project Updates:
- Update requirements.txt with new dependencies


kga245:
Here is a loom so you can see and hear my rationale for this approach. https://www.loom.com/share/0872c41700c840d6b42c3d2d8e8191b8

kga245:
Changed the readme to better decribe the configs update.

ElishaKay:
Interesting 

What's your take on customisation around the vector store?

For example, the Langchain Vector Store object supports filtering search results based on metadata (like user_id, project_id, task_id, domain_id...)

Would be cool if a single Vector Store would be able to support a multi_user, multi_projects server

kga245:
@ElishaKay 

Don't merge. Don't bother testing either. I am going to git hard reset this one. I discovered a number of changes to core functionality I don't think we should be comfortable making. I need to package this better. So back to the drawing board I go! But it's helpful to see your notes. I'm going to draw up specs first this time. Do you want me to post those to the wiki first? Issues? Discussions? Projects? I am comfortable making a recommendation too if you like.

PR #1059, opened 2025-01-03
Remove server log.txt

kga245:
small change. tidying up a line of code that was automatically creating a file server_log.txt at the root of the application. that was getting committed too so I added the necessary lines to gitignore so we're not constantly loading an empty server_log.txt file when we pull.


kga245:
just some housekeeping.

assafelovic (approved):
Thanks!

PR #1060, opened 2025-01-04
Add Litellm to supported providers in GenericLLMProvider

moritalous:
Add Litellm to supported providers in GenericLLMProvider.

Using Litellm has the following advantages:
Call 100+ LLMs using the OpenAI Input/Output Format

https://docs.litellm.ai/


assafelovic (approved):
This is a great addition thanks @moritalous 


FYI @ishaan-jaff

assafelovic:
@ishaan-jaff We're finally pushing Litellm. How about we announce it on socials?

PR #1070, opened 2025-01-08
preserve user-specified report source instead of config default

alesaccoia:
The report_source parameter was being incorrectly overridden by the config  default value due to Python's 'or' operator behavior with truthy values.  Changed the logic to explicitly check for the passed parameter first before  falling back to config values.


lmontin:
I tested the change and it worked. ie when using the nextjs ( port 3000 ) interface my preferences selection for report source My Documents ( local ) was respected

alesaccoia:
Cool. Just added some logging for Local mode in the last commit

assafelovic (approved):
This is super helpful and important bug fix thank you @alesaccoia !

PR #1116, opened 2025-02-06
Fixed human in the loop

electricjimi:
Since self.websocket is wrapped inside a CustomLogsHandler in websocket_manager.py, the call to

```python

response = await self.websocket.receive_text()
```

was always throwing an exception.


electricjimi:
@assafelovic please look into this, may be useful for those using multi-agent research mode

assafelovic (approved):
Thank you for this super important fix!

PR #1123, opened 2025-02-10
fix: update directory nextjs example in langgraph.md

fibonacci998:
Update wrong directory nextjs in current example in langgraph.md


assafelovic (approved):
Thank you!

PR #1135, opened 2025-02-12
missing langchain-google-genai for Gemini provider

denispol:
Using Gemini as provider was broken due to missing dependency.
Added langchain-google-genai to the requirements.txt, as Gemini provider cannot be used without it

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,6 +11,7 @@ python-multipart
 markdown
 langchain
 langchain_community
+langchain-google-genai
 langchain-openai
 langchain-ollama
 langgraph


ElishaKay:
Welcome @denispol 

This is intentional, since we'd like to keep the requirements.txt as lean as possible.
We ask that users install llm-specific dependencies.

What we may want to do is add better error handling around this case.
For example, if a python package of an LLM is missing and the user tries to run a report, we can have a more meaningful error message, like:

"To keep the requirements.txt as lean as possible, we ask that users install llm-specific dependencies."

assafelovic:
@denispol as @ElishaKay stated we want to keep the library lean and light so any additional package other than default requires manually installing it. THanks!

PR #1143, opened 2025-02-12
refactor: get text from soup

ewgdg:
refactor for the soup utils


assafelovic (approved):
This is a great addition!

PR #1150, opened 2025-02-14
added helpful optional configs for dockerized hosting

kga245:
added helpful optional configs for dockerized hosting
Update .env.example


assafelovic (approved):
Thanks @kga245 maybe a few around GPTR config could also help

Issue #1047, opened 2024-12-27
Tavily - To Many Requests


jovan-actdigital:
**Describe the bug**
Error: 429 Client Error: Too Many Requests for url: https://api.tavily.com/search. Failed fetching sources. Resulting in empty response.

Then the output for any query is empty.

**To Reproduce**
1.Clone the project
2.Install requirements.txt
3.Put in env tavily  and openai key 
4.Run the program
5. Put query  "Stock analysis on NVIDIA"

**Expected behavior**
Browse the web and return report.

assafelovic:
@jovan-actdigital maybe you've reached your limit? Check it in https://app.tavily.com

jovan-actdigital:
When i  start the app with docker it runs fine, do you know why is that?

iliyandobrev:
I have the exact same problem. This only happens if I run it from FastAPI endpoint. Please suggest a fix if possible.

assafelovic:
Closing as part of a long-tail backlog cleanup — this issue has been open for **over a year** with no recent activity. This is not a judgment on its validity; we're clearing the aged backlog so the team can focus on current, reproducible issues.

**If this is still relevant on the latest version, please reopen (or comment) with up-to-date details — version, config, and a repro — and we'll gladly take another look.** Thank you for the report! 🙏

PR #1170, opened 2025-02-19
add async nodriver scrapper

ewgdg:
Add support for async scrapper.

Add a new scrapper based on [zendriver](https://github.com/stephanlensky/zendriver) (fork of [nodriver](https://github.com/ultrafunkamsterdam/nodriver)), which is fast and async-first for web scrapping. 
This enables rooms for better performance.

Additionally, add a new configs for throttling scraper concurrency. `MAX_SCRAPPER_WORKERS`. 
This might helps complying with the rate limits of some services/websites. see also https://github.com/assafelovic/gpt-researcher/issues/1047.


ElishaKay:
Looks good @ewgdg 

Looks like there are some conflicts to resolve before merge
May the force be with you 🙏

ewgdg:
I tried bs, browser, tavily on the react frontend.

one issue I noticed for the browser slow scrapping is that sometimes the websocket can be prematurely closed right before generating report. I suspect this is caused by the frontend sending `newSocket.send('ping');` too aggressively. The server doesn't seem to expect the `ping` message. commenting out the ping or decrease the frequency seems to resolve the issue. but I am not sure if this is the root cause.

ewgdg:
I kept max_http_sessions only inside the nodriver scraper since other scrapers don’t use it.

ElishaKay (approved):

ElishaKay:
Good stuff @ewgdg  

Welcome to the Git Tree 🏕️

PR #1177, opened 2025-02-21
Add FireCrawl scaper

kongacute:
Add support for FireCrawl scaper for page content and page title.
It should be support self host or official service from FireCrawl by set enviroment variables `FIRECRAWL_SERVER_URL` and `FIRECRAWL_API_KEY`.
And set `SCRAPER=firecrawl` to use FireCrawl scaper.

Already tested with selft hosted FireCrawl.

### DIFF
--- a/docs/docs/gpt-researcher/gptr/scraping.md
+++ b/docs/docs/gpt-researcher/gptr/scraping.md
@@ -16,10 +16,15 @@ You can choose your preferred scraping method by setting the `SCRAPER` environme
    export SCRAPER="browser"
    ```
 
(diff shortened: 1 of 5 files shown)


assafelovic (changes requested):
This is great we love firecrawl! Can you please also add documentation here: https://github.com/assafelovic/gpt-researcher/blob/master/docs/docs/gpt-researcher/gptr/scraping.md

kongacute:
Hi @assafelovic , I just updated the docs. Could you take a look at it?

assafelovic (approved):
This is great thank you!

PR #1178, opened 2025-02-21
Fix: ArxivScraper causing error when scraping the link.

new-player:
Screenshot 2025-02-21 at 12 47 39 PM

On setting RETRIEVER=arxiv, an error comes and content is not retrieved.
Reference for fix: https://python.langchain.com/docs/integrations/retrievers/arxiv/#usage


assafelovic (approved):
Great fix thank you!

PR #1192, opened 2025-02-24
Fix Docker Build Failure: Updated `combined_query` in `DeepRsearchSkill.run()` to Handle Backslashes in F-Strings

monolok:
I changed the combined_query in the run() method of DeepRsearchSkill class to solve a bug with backlashes in f-strings that was causing the building of the docker image to fail - now the build work.


ElishaKay (approved):

ElishaKay:
awesome
thanks @monolok!

Issue #1185, opened 2025-02-23
Add base_url param support for LLM / VLM


d6l8:
**Is your feature request related to a problem? Please describe.**
I need to use OpenAI models with the `base_url` param, but simply putting `OPENAI_API_KEY` in the `.env` file doesn't work because the function `create_chat_completion` in the `gpt-researcher\gpt_researcher\utils\llm.py` doesn't support this param. And your code for the OpenAI embedding model supports this param.
```
_embeddings = OpenAIEmbeddings(
    model=model,
    openai_api_key=os.getenv("OPENAI_API_KEY", "custom"),
    openai_api_base=os.getenv(
        "OPENAI_BASE_URL", "http://localhost:1234/v1"
    ),  # default for lmstudio
```

**Describe the solution you'd like**
Add support for the `base_url` param in the func `create_chat_completion`, which reads the `base_url` param from the `.env` file like above.

PR #1198, opened 2025-02-24
Added support for base_url param in create_chat_completions for OpenAI Provider

gaurav3247:
Resolved issue #1185 by checking for the base url in the environment variables and adding it to the arguments sent to llm provider.


assafelovic (approved):
Thank you!

PR #1208, opened 2025-02-26
Add language requirement to resource report prompt

hurxxxx:
Other reports already support multiple languages, but it seems that the Resource Report was missing this feature. This PR adds multi-language support to ensure consistency


ElishaKay (approved):

ElishaKay:
Good stuff @hurxxxx 

Welcome to the Git Tree 🏕️

PR #1215, opened 2025-02-27
Fix: image relevance filtering

czakop:
I found a strange thing in the current image filtering logic, that seems to ignore images with the highest "relevance score" 4, and also performs unnecessary steps.

Maybe I'm missing something here, but I wouldn't expect this behavior so implemented a fix, as well as some type hinting clean up in the affected files.


assafelovic (approved):
I like this approach. First of all good call on missing out score 4! Secondly, the original approach was to allow for smaller images as well to provide a wide range of options. But, I like the simplicity of this approach, so happy to move forward with it. If we hear out from the community about this change we might consider reverting. Thank you!

PR #1219, opened 2025-03-01
perf: load balance nodriver browser scraper tabs

ewgdg:
evenly distribute the tabs to different browser instances to improve performance.


assafelovic (approved):
Thank you!

PR #1220, opened 2025-03-01
fix: unblock web socket message queue

ewgdg:
a long running task might block the web socket from handling new incoming messages, and thus the messages will be accumulated in the queue. eventually the queue can be full and lead to unexpected behavior.

This fix puts long running tasks to background and discard any new request except for the ping message.


assafelovic (approved):
This is great and love the handling across the entire app. Great job!

PR #1238, opened 2025-03-07
fix: resolve failing test in tests/test_logging_output.py

grapestore:
Internal package’s WebSocket implementation, if bool is not defined, Python falls back to calling len. In our case, len returns the length of self.scope, which has been updated to always include at least the key "type": "websocket" to match the internal package implementation.

Additionally, for test stability and to verify the existence of the TestWebSocket instance, I have overridden bool in TestWebSocket class.


assafelovic (approved):
This is great thank you!

PR #1241, opened 2025-03-08
feat: add open-ai compatible endpoints

ewgdg:
This enables integrating this agent into UIs that is built for open-ai compatible REST APIs.
e.g. [open-webui](https://github.com/open-webui/open-webui).


backend/chat/chat.py
```diff
@@ -82,20 +91,12 @@ def _process_document(self, report):
...
         """Chat with React Agent"""
-        message = f"""
-         You are GPT Researcher, a autonomous research agent created by an open source community at https://github.com/assafelovic/gpt-researcher, homepage: https://gptr.dev. 
-         To learn more about GPT Researcher you can suggest to check out: https://docs.gptr.dev.
-         
-         This is a chat message between the user and you: GPT Researcher. 
-         The chat is about a research reports that you created. Answer based on the given context and report.
-         You must include citations to your answer based on the report.
-         
-         Report: {self.report}
```
ewgdg, line 95: I removed this report from the prompt, given that the report is added to the vector store. but just want to confirm that adding the full report to the prompt is not intended.

assafelovic:
Hey @ewgdg thanks can you share more info about the motivation for this? What does it solve for you?

ewgdg:
The main motivation is to be able to use it in Open-WebUI.
The current websocket endpoint is difficult to be integrated into Open-WebUI. Each time you disconnect the websocket, you are not able to continue the conversation. So you cannot review a previous conversation and continue it.

But I can move the bridge to Open-WebUI side. Just need to keep the fix to `backend/chat/chat.py`. Maybe this will be easier.

ElishaKay:
Sup @ewgdg 

Does it solve this one?

https://github.com/assafelovic/gpt-researcher/issues/1293

PR #1242, opened 2025-03-08
LangGraph x GPTR Deep Research

assafelovic:


assafelovic:
Triage note — this looks **superseded**, worth a decision rather than more waiting.

Opened 2025-03; `main` has since grown deep research independently: `gpt_researcher/skills/deep_research.py`, the `deep_agents/` package, and `backend/report_type/deep_research/`. The 4,752-line diff across 29 files no longer applies.

Unless there's something in here the current implementation lacks, closing it is probably kinder than leaving it open — it's one of the two oldest PRs in the queue and reads as live work to anyone browsing. Happy to diff it against the current implementation to see whether anything is genuinely missing before you decide.

PR #1257, opened 2025-03-12
fix: truncating task file names for long queries

pavlus007:
Long queries still make task log file paths too long for Windows to handle (255) which causes frontend app to throw errors and crash research task at the start (tested locally).  Current truncating mechanism doesn't include real length for path to project, it just leaves 10 chars for safety margin


assafelovic (approved):
Awesome, thank you!

Issue #1216, opened 2025-02-28
Multi Agents Report Calls "gpt-4o" Instead of Local Ollama Model Before Generating PDF/DOCX


Aerobit:
**Describe the bug**
When generating a Multi-Agent Report, GPT Researcher uses the correct Ollama model initially, but right before the report finishes (before "View as PDF" and "Download DOCX" appear), it makes an incorrect call to "gpt-4o". This results in a 404 Not Found error because "gpt-4o" is not available in Ollama.

**To Reproduce**
1. Install GPT-Researcher & Ollama in Docker
2. Configure any local Ollama model in .env file for Fast, Smart and Strategic models
3. Choose  Report Type=Multi Agents Report,  Report Source=Internet, Tone=Objective

**Expected behavior**
 -The Multi-Agent Report should only use the configured Ollama model 
 -No request should be made to "gpt-4o".

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Docker Install

**Additional context**
2025-02-28 00:33:27 gpt-researcher-1  | INFO:     [08:33:27] ✍️ Writing report for 'Stock analysis on Bircoin'...
2025-02-28 00:33:35 gpt-researcher-1  | 2025-02-28 08:33:35,845 - httpx - INFO - HTTP Request: POST http://host.docker.internal:11434/api/chat "HTTP/1.1 200 OK"
2025-02-28 00:34:13 ollama-1          | [GIN] 2025/02/28 - 08:34:13 | 200 |  46.03604018s |      172.18.0.1 | POST     "/api/chat"
2025-02-28 00:34:13 gpt-researcher-1  | INFO:     [08:34:13] 📝 Report written for 'Stock analysis on Bircoin'
2025-02-28 00:34:13 ollama-1          | [GIN] 2025/02/28 - 08:34:13 | 404 |       953.6µs |      172.18.0.1 | POST     "/api/chat"
2025-02-28 00:34:13 gpt-researcher-1  | 2025-02-28 08:34:13,338 - httpx - INFO - HTTP Request: POST http://host.docker.internal:11434/api/chat "HTTP/1.1 404 Not Found"
2025-02-28 00:34:13 gpt-researcher-1  | 2025-02-28 08:34:13.339 | ERROR    | multi_agents.agents.utils.llms:call_model:49 - Error in calling model: model "gpt-4o" not found, try pulling it first (status code: 404)

ElishaKay:
Sup @Aerobit 

[...]

PR #1258, opened 2025-03-13
fix: replace gpt-4o hardcode model in multi-agent with env var.

galenyip:
related to issue mentioned in https://github.com/assafelovic/gpt-researcher/issues/1216

and fix according to the suggestion:
https://github.com/assafelovic/gpt-researcher/issues/1216#issuecomment-2692203134


assafelovic (approved):
Long overdue thanks so much for pushing it!

PR #1280, opened 2025-03-19
feat: Adds support for gigachat embeddings to the gigachat LLM add-on.

Rai220:
LLM already works, adds embeddings from GigaChat team.


assafelovic (approved):
This is awesome thank you

Issue #1232, opened 2025-03-04
gpt-researcher npm package does not support https host


vikram-wald:
Deployed the gpt-reseacher in google cloud run
the server is running in https protocol
I have integrated websocket using gpt-researcher npm package, and it is not working since it uses `ws://..` instead of `wss://..`
Tried connecting using `wss://..` , it works
We need to support https also in the npm package

ElishaKay:
Welcome @vikram-wald 

This a great point actually!

Some thoughts on what we need to achieve http / https for the server and npm package:

A) The python server exposes an http get request route when a report completes (where the report id is in the url itself

B) we'll want to add a new HTTP post request route to the python server - this route should either return the request id or the completed report (the http connection may time out depending on the client) - therefore we can add a parameter to the post request to either just fetch the report id or wait for the report

C) we'll want to add the http option to the [npm package code here](https://github.com/assafelovic/gpt-researcher/blob/master/docs/npm/index.js)

Green light for the PR if someone from the community wants to take it

gaurav3247:
@vikram-wald http support has been added to the npm package. Will update the docs soon

assafelovic:
A contributor noted that http/https (and wss) support was since added to the npm package. Closing as resolved — please reopen if you still see `ws://` being forced on an https host on the latest version. Thanks!

PR #1283, opened 2025-03-21
Added post and get http endpoints to server and updated npm package to allow sending http requests

gaurav3247:
Resolving issue https://github.com/assafelovic/gpt-researcher/issues/1232
The server now accepts post requests for generating reports. The execute_in_background flag can be used to wait for the report to finish generating and return the report, or just return the report_id to then fetch the report later using a get request.

The sendMessage function in npm package now accepts useHTTP and execute_in_background params to decide whether to use websockets or HTTP and whether to wait for the report to finish. The npm package also has a getReport function which accepts the report_id and retrieves the report.

### DIFF
--- a/backend/report_type/basic_report/basic_report.py
+++ b/backend/report_type/basic_report/basic_report.py
@@ -29,9 +29,8 @@ class BasicReport:
         self.websocket = websocket
         self.headers = headers or {}
 
-    async def run(self):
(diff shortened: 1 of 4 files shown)


assafelovic (approved):
Thank you this is great can you please update in the issues? You rock, keep em' coming!

PR #1285, opened 2025-03-21
Fix AttributeError in PublisherAgent when processing string values

LavX:
## Issue
When running the multi-agent system, an error occurs in the `PublisherAgent` class when trying to process research data that contains string values instead of dictionaries:

```
AttributeError: 'str' object has no attribute 'items'
```

This happens because the `generate_layout` method assumes all items in `research_data` are dictionaries with an `.items()` method, but sometimes they can be strings.

## Solution
Modified the `generate_layout` method in `PublisherAgent` to handle both dictionary and string values in the research data:

1. Added type checking with `isinstance(subheader, dict)` to handle different data types
2. Changed from a nested list comprehension to a more explicit loop structure
3. Added proper handling for string values in the research data
4. Added default empty values for collections to prevent errors when keys are missing

## Changes
- Updated `multi_agents/agents/publisher.py` to make the `generate_layout` method more robust

### DIFF
--- a/multi_agents/agents/publisher.py
+++ b/multi_agents/agents/publisher.py
@@ -20,11 +20,19 @@ class PublisherAgent:
         return layout
 
     def generate_layout(self, research_state: dict):
-        sections = '\n\n'.join(f"{value}"
-                                 for subheader in research_state.get("research_data")
-                                 for key, value in subheader.items())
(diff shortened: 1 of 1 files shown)


assafelovic (approved):
Love this thank you

PR #1350, opened 2025-04-22
Strip spaces from "self.output_dir" to prevent exception when writing…

ivan-latunov:
… files


ivanlatunov:
Removed spaces from self.output_dir to avoid error when writing file

assafelovic (approved):
thanks!

Issue #1360, opened 2025-04-28
verbosity flag should control all print statements


tClelford:
**Describe the bug**
When passing `verbose=False` to `GPTResearcher`, the report is still printed to stdout as it is generated.  When running this as a component in a production system, this fills the logs with report text, obscuring more useful information. 

**To Reproduce**
1. Create an instance of `GPTResearcher` and pass the param `verbose=False`.
2. Call `write_report`
3. Observe green text streaming to stdout.

**Expected behavior**
When verbose is set to false, only logs are written.

**Additional context**
[This](https://github.com/assafelovic/gpt-researcher/blob/bfea7d2e7ea46c2b3a57ad3d946d54644f769790/gpt_researcher/llm_provider/generic/base.py#L251) is the function that needs to change. If we create a `print_stream:bool=True` parameter and populate it from `GPTResearcher.verbose` then we should get my desired behaviour without changing the default behaviour.

[This issue](https://github.com/assafelovic/gpt-researcher/issues/472) seems related, not sure if this bit was missed when fixing it or if it's just a regression

PR #1362, opened 2025-04-28
Respect my verbosity! 

tClelford:
This fixes [the issue](https://github.com/assafelovic/gpt-researcher/issues/1360) I raised earlier. After going through the codebase, it seemed neater to pop the verbose property on `Config` and populate it. This does create a potential bug where somebody could set `LLM_KWARGS.verbose` in their config file and then it could be overridden by the parameter passed to an instance of the GPT researcher object, but I feel like this gives more granular control than always deferring to the config file.


assafelovic (approved):
thank you this was much needed to fix

PR #1363, opened 2025-04-28
fix firecrawl error: 'ScrapeResponse' object is not subscriptable

tvoilex:
https://docs.firecrawl.dev/features/scrape


assafelovic (approved):
amazing thank you

PR #1454, opened 2025-07-02
Fix zero subtopic error

Yuxuan1998:
**Problem Description**
When generating subtopic reports for the detailed report, the following error was encountered:
`Exception in parsing subtopics : Connection error`
As a result, neither the table of contents nor the subtopic reports are generated in the final output report.

**Solution**
- Changed the blocking `invoke` call to the asynchronous `ainvoke` to avoid blocking the event loop.
- Added additional logging for improved debugging and traceability.


assafelovic (approved):
thank you !

PR #1461, opened 2025-07-09
bug_fix: vectorstore_filter not passed

viboognesh:
**Describe the bug**
The vectorstore_filter parameter passed in GPTResearcher class is not passed the VectorstoreCompressor and VectorStoreWrapper class while conduct research.

**To Reproduce**
In `gptresearcher/context/compression.py -> VectorStoreCompressor` Make this change to async_get_context
```
async def async_get_context(self, query, max_results=5, cost_callback=None):
    print(f"DEBUG: VectorstoreCompressor.async_get_context using filter: {self.filter}")
    results = await self.vector_store.asimilarity_search(query=query, k=max_results, filter=self.filter)
    print(f"DEBUG: VectorstoreCompressor.async_get_context raw results (first 2 metadata): {[r.metadata for r in results[:2]]}")
    return self.prompt_family.pretty_print_docs(results)
```

In `gptresearcher/vector_store/vector_store.py -> VectorStoreWrapper` Make this change to asimilarity_search

```
async def asimilarity_search(self, query, k, filter):
      print(f"DEBUG: VectorStoreWrapper.asimilarity_search received filter: {filter}")
      results = await self.vector_store.asimilarity_search(query=query, k=k, filter=filter)
      print(f"DEBUG: VectorStoreWrapper.asimilarity_search results (first 2 metadata): {[r.metadata for r in results[:2]]}")
      return results
```

The filter is not passed here
```
DEBUG: VectorstoreCompressor.async_get_context using filter: None
DEBUG: VectorStoreWrapper.asimilarity_search received filter: None
DEBUG: VectorstoreCompressor.async_get_context using filter: None
DEBUG: VectorStoreWrapper.asimilarity_search received filter: None
DEBUG: VectorstoreCompressor.async_get_context using filter: None
DEBUG: VectorStoreWrapper.asimilarity_search received filter: None
DEBUG: VectorstoreCompressor.async_get_context using filter: None
DEBUG: VectorStoreWrapper.asimilarity_search received filter: None
DEBUG: VectorStoreWrapper.asimilarity_search results (first 2 metadata): [{'fileid': 'jun28'}, {'fileid': 'jun26'}]
DEBUG: VectorstoreCompressor.async_get_context raw results (first 2 metadata): [{'fileid': 'jun28'}, {'fileid': 'jun26'}]
DEBUG: VectorStoreWrapper.asimilarity_search results (first 2 metadata): [{'fileid': 'jun28'}, {'fileid': 'jun26'}]
[...]

PR #1463, opened 2025-07-11
Bug fix - filter argument was not properly passed on

coreation:
Hello, 

I noticed the filter value in the VectorstoreCompressor was None and tracked the issue back to this place. Just adding the name of the argument fixes the issue.


coreation:
duplicate of #https://github.com/assafelovic/gpt-researcher/pull/1461

PR #1465, opened 2025-07-16
Update config.md

sdisaacson:
Removed mentions of config_file param because it doesn't exist.

Corrected the file to mention the config_path parameter and added a sample json file as well as a sample CLI command starting GPT-Researcher and specifying a specific config.json file should be used.

Do with this what you will, config_file should be changed to config_path at a bare minimum.


assafelovic (approved):
thank you !

PR #1511, opened 2025-09-27
feat: Add PWA capabilities to the frontend

TheSpaceGod:
I asked Google Jules to add PWA support to the gpt-researcher app so it can be used outside of a browser and feel more like an "app" (even though it uses the browser under the hood).  To be clear, I am not a web developer and have no idea if this code properly works or not.  Jules said it tested the code and I figure it would be worth sharing.

This commit introduces Progressive Web App (PWA) capabilities to the Next.js frontend application.

- Installed and configured the `@ducanh2912/next-pwa` package to manage PWA functionalities.
- Updated `next.config.mjs` to integrate the PWA plugin, enabling service worker generation and registration.
- Replaced the basic service worker (`sw.js`) with a more robust version for better caching and offline support.
- Removed the manual service worker registration script from `layout.tsx` as it is now handled by the `next-pwa` plugin.
- Fixed several ESLint errors to ensure a successful build.


assafelovic (approved):
This is great, thank you!

PR #1541, opened 2025-10-30
Update got-researcher

delormeca:


assafelovic:
@delormeca can you please share some info about the updates and changes you did?

PR #1553, opened 2025-11-12
Add output format control flags to CLI for selective report generation

0xsalt:
# what it does
Enhance CLI with --no-pdf and --no-docx flags to allow users to skip specific output formats. 

# default behavior
By default, all three formats (markdown, PDF, DOCX) are generated. 

# benefit
This provides flexibility for users who only need specific output formats without modifying code.

# Examples:
- python cli.py "What is Claude?" --report_type research_report  //generates .md, .docx & .pdf outputs 
- python cli.py "What is Claude?" --report_type research_report --no-pdf  //generates .md & .docx outputs 
- python cli.py "What is Claude?" --report_type research_report --no-pdf --no-docx  //generates .md output only

### DIFF
--- a/cli.py
+++ b/cli.py
@@ -17,8 +17,9 @@ import os
 from dotenv import load_dotenv
 
 from gpt_researcher import GPTResearcher
-from gpt_researcher.utils.enum import ReportType, Tone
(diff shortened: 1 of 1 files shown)


assafelovic (approved):
You're on fire @0xsalt really valuable additions

PR #1630, opened 2026-02-17
Attempt to load config path from env var if no path provided at runtime

GeorgelPreput:
When deploying with Docker, currently it's cumbersome to provide a custom configuration file. The way to do it with no code changes would be to create a `default.py` dictionary and override-mount the existing file in the `/gpt_researcher/config/variables` folder.

This is not optimal, especially since there's already a robust way to load a JSON configuration file when instantiating GPT Researcher from a Python script. Therefore, I propose this minimal alteration:

If no `config_path` is provided at runtime, try loading it from the environment variables.


assafelovic (approved):
Great fix, love it, thank you!

PR #1637, opened 2026-02-20
fix: handle list context in deep research to prevent AttributeError

technot80:
- count_words() now handles both string and list inputs defensively
- context is normalized to string when returned from nested research
- Prevents 'list' object has no attribute 'split' error during recursive deep research when LLM returns unexpected types


assafelovic (approved):
This is great, thank you and good catch!

Issue #1667, opened 2026-03-09
Detailed report with MCP results results in hashing error


GeorgelPreput:
**Describe the bug**
When generating a Detailed Report, while having MCP servers configured (and getting some results from them), the report generation fails when deduplicating context due to MCPs returning context in the form of dictionaries. The `detailed_report.py` code attempts to do `set(...)` of items in both the global context and subtopic_assistant context. If there's any MCP output in either of them, report generation will fail.

**To Reproduce**
1. In the UI, go to Preferences
2. Set Report Type to Detailed
3. Enable MCPs
4. Configure some MCPs that will get used in the report
5. Generate a report and watch for the stack trace below

**Expected behavior**
A report is generated successfully without the hashing error

**Screenshots**
```
gpt-researcher  | 2026-03-09 12:55:32,548 - server.server_utils - ERROR - Error running task: unhashable type: 'dict'
gpt-researcher  | Traceback (most recent call last):
gpt-researcher  |   File "/usr/src/app/backend/server/server_utils.py", line 331, in safe_run
gpt-researcher  |     await awaitable
gpt-researcher  |   File "/usr/src/app/backend/server/server_utils.py", line 158, in handle_start_command
gpt-researcher  |     report = await manager.start_streaming(
gpt-researcher  |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
gpt-researcher  |     ...<12 lines>...
gpt-researcher  |     )
gpt-researcher  |     ^
gpt-researcher  |   File "/usr/src/app/backend/server/websocket_manager.py", line 106, in start_streaming
gpt-researcher  |     report = await run_agent(
gpt-researcher  |              ^^^^^^^^^^^^^^^^
gpt-researcher  |     ...<3 lines>...
gpt-researcher  |     )
gpt-researcher  |     ^
gpt-researcher  |   File "/usr/src/app/backend/server/websocket_manager.py", line 162, in run_agent
gpt-researcher  |     report = await researcher.run()
gpt-researcher  |              ^^^^^^^^^^^^^^^^^^^^^^
gpt-researcher  |   File "/usr/src/app/backend/report_type/detailed_report/detailed_report.py", line 82, in run
gpt-researcher  |     _, report_body = await self._generate_subtopic_reports(subtopics)
gpt-researcher  |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1668, opened 2026-03-09
Fix: Add context normalization for dict and string formats

GeorgelPreput:
Due to the issue in #1667 I propose that in the Detailed Report generation, both dict context items from MCP, and strings from everywhere else are normalized by converting them to a consistent hashable string format before deduplication.


assafelovic (approved):

assafelovic:
@GeorgelPreput thanks for this, due to some early merges of updated code, can you take a look at the conflict and fix? Thanks! Love this

GeorgelPreput:
Fixed the conflicts, could you have another look?

assafelovic (approved):
i love this thank you

Issue #1386, opened 2025-05-06
Deep research - request too large for gpt-4.1


JoeRoddy:
**Describe the bug**
"deep" mode throws an error when writing report, context too large for gpt-4.1 using python package

```
Error in generate_report: Error code: 429 - {'error': {'message': 'Request too large for gpt-4.1 in organization org-*** on tokens per min (TPM): Limit 30000, Requested 49372. The input or output tokens must be reduced in order to run successfully. Visit https://platform.openai.com/account/rate-limits to learn more.', 'type': 'tokens', 'param': None, 'code': 'rate_limit_exceeded'}}
```

**To Reproduce**
- Set `OPENAI_API_KEY` env var
- Set `TAVILY_API_KEY` env var
- run gpt researcher in deep mode:
```
researcher = GPTResearcher(query, "deep")
research_result = await researcher.conduct_research()
report = await researcher.write_report()
```

**Expected behavior**
[The docs](https://docs.gptr.dev/docs/gpt-researcher/gptr/deep_research) seem to imply it would default to o3-mini when in deep mode, as thats the only model that's referenced, but it seems to use gpt-4.1 and hit the context limit.

**Desktop (please complete the following information):**
 - OS: MacOS - ran as python package

JoeRoddy:
I'm able to fix this by setting `SMART_LLM=openai:o4-mini` in my .env, but am wondering if I'm potentially doing something else wrong.

maxwelljmckee:
@JoeRoddy I'm having the same issue. `gpt-4.1` consistently returns an empty report. I've had success generating `deep` research reports with both `o4-mini` and `gpt-4.1-mini`, with the latter being significantly higher quality in my opinion. However, I'd still like to get a result with the higher end `gpt-4.1` model to compare with `gpt-4.1-mini`.

I've also got an [ongoing thread](https://github.com/assafelovic/gpt-researcher/issues/1378) in Issues with @assafelovic regarding this issue. Let's keep in touch and see if we can resolve this!!

apunkt:
gpt-4.1 has 30k TPM,
o4-mini has 200k TPM.

This was causing my trouble.
Check: https://platform.openai.com/settings/organization/limits

apunkt:
..."tokens per min (TPM): Limit 30000, Requested 49372."

You are overwhelming the model, hitting the limit, OpenAI refuses to answer, the research must fail.
[...]

Issue #1378, opened 2025-05-02
`researcher.write_report()` yields incomplete/inconsistent results


maxwelljmckee:
**Describe the bug**
While experimenting with GPTR, I'm finding that my report generation outputs are inconsistent. About half the time, the report appears to be cut off mid-generation, sometimes even ending mid-sentence. I am having difficulty determining whether this issue is internal to GPTR, or an issue with my genai provider.

**Environment Setup**
I am using `langchain_google_genai` with `tavily` search as the retriever. Here is a bit of my environment setup:
```
GOOGLE_API_KEY="[secret]"
FAST_LLM="google_genai:gemini-2.0-flash-lite"
SMART_LLM="google_genai:gemini-2.0-flash"
STRATEGIC_LLM="google_genai:gemini-2.0-flash"
# SMART_LLM="google_genai:gemini-2.5-flash-preview-04-17"
# STRATEGIC_LLM="google_genai:gemini-2.5-flash-preview-04-17"
# SMART_LLM="google_genai:gemini-2.5-pro-preview-04-17"
# STRATEGIC_LLM="google_genai:gemini-2.5-pro-preview-04-17"
EMBEDDING="google_genai:models/text-embedding-004"
TOTAL_WORDS=5000
MAX_SUBTOPICS=5
RETRIEVER="tavily, duckduckgo"
SCRAPER="tavily_extract"
```
As you can see, I've tried several different google genai models. The gemini-2.0 models tend to yield better results that 2.5, but I have experienced incomplete outputs with all of the above models.

**Expected behavior**
Expected behavior is that the report generation is completed before the GPTR process is terminated

**Actual behavior**
Report is cut off mid-section, mid-sentence

**Screenshots**

**Desktop (please complete the following information):**
 - OS: MacOS 15.3.2
 - Running the report generation locally as a python script in a terminal

```py
import json
import os
import asyncio
from datetime import datetime
from gpt_researcher import GPTResearcher
from prompts import document_analysis_prompt


async def gpt_researcher(query: str) -> dict:
    researcher = GPTResearcher(
        query,
        report_type='deep',
        report_source='tavily-search',
    )
    research_result = await researcher.conduct_research()
    report = await researcher.write_report()

    completion = {
        'report': report,
    }

    output_dir = 'outputs'
    os.makedirs(output_dir, exist_ok=True)  # Ensure the directory exists
    output_file_path = os.path.join(
[...]

PR #1671, opened 2026-03-12
Add checkpoint persistence and retry logic for deep research mode

JasonBroderick:
## Summary

Deep research mode stores all intermediate data (scraped content, sub-reports, citations, sources) only in memory. If the final `generate_report()` call fails — due to rate limits, token limits, or any other error — **all research work is lost with no recovery path**. This PR adds crash recovery and resilience as a first-class feature.

- **Checkpoint persistence**: After `conduct_research()` completes, researcher state (context, visited URLs, sources, costs, config) is serialized to a JSON file. On the next run with the same query, research is skipped and only report generation retries.
- **Retry with exponential backoff**: `generate_report()` retries on 429 rate limit errors with 30s/60s/120s delays (configurable max retries).
- **Context chunking fallback**: When token limits are exceeded, context is automatically split into ~4k word chunks, partial reports generated, then synthesized into a final report.
- **Model fallback**: New `SMART_LLM_FALLBACK` config — if the primary model exhausts retries, a cheaper/smaller model is attempted before giving up.

## New Configuration

| Variable | Default | Purpose |
|----------|---------|---------|
| `ENABLE_CHECKPOINTS` | `true` | Enable/disable checkpoint persistence |
| `CHECKPOINT_DIR` | `~/.gpt-researcher/checkpoints` | Where to store checkpoints |
| `CHECKPOINT_MAX_AGE_HOURS` | `168` (7 days) | Auto-cleanup threshold |
| `SMART_LLM_FALLBACK` | `None` | Fallback model for report generation |
| `REPORT_GENERATION_RETRIES` | `3` | Max retries on rate limit errors |

## Files Changed

| File | Change |
|------|--------|
| `gpt_researcher/utils/checkpoint.py` | **New** — checkpoint manager (save/find/restore/cleanup) |
| `gpt_researcher/actions/report_generation.py` | Retry + backoff + chunking + model fallback |
| `gpt_researcher/skills/deep_research.py` | Checkpoint save/restore in `run()` |
| `gpt_researcher/agent.py` | Update checkpoint status on report success/failure |
| `gpt_researcher/config/variables/base.py` | New config type hints |
| `gpt_researcher/config/variables/default.py` | New config defaults |
| `gpt_researcher/config/config.py` | Parse fallback LLM config |

## Motivation

Addresses:
- **#1386** — Rate limit (429) errors during report generation cause total research loss
- **#1378** — Token limit exceeded errors fail silently with no recovery

## Design Decisions

- **JSON format**: Human-readable, one file per research run, easy to inspect/debug
- **`raw_content` excluded** from checkpoints: Too large; the synthesized `context` is sufficient for report generation
- **Atomic file writes**: Write to temp file, then `os.replace()` for thread safety
- **Query matching**: Slugified comparison (case-insensitive, punctuation-stripped) so minor query variations still match
- **Backward compatible**: All new features are opt-in via config. Existing behavior is unchanged when `ENABLE_CHECKPOINTS=false`.

## Test plan

- [ ] Happy path: Run deep research, verify checkpoint written + cleaned up on success
- [ ] Report failure recovery: Trigger a 429 or token limit error, verify checkpoint preserved with `report_failed` status, re-run same query and confirm research is skipped
- [ ] Model fallback: Set `SMART_LLM_FALLBACK`, verify fallback fires after primary fails
- [ ] Retry logic: Verify 429 retry with backoff appears in logs
- [ ] Stale cleanup: Verify checkpoints older than `CHECKPOINT_MAX_AGE_HOURS` are auto-removed
- [ ] Backward compat: Run with `ENABLE_CHECKPOINTS=false`, verify no checkpoint files created

### DIFF
--- a/.agents/skills/gpt-researcher/SKILL.md
+++ b/.agents/skills/gpt-researcher/SKILL.md
@@ -1,0 +1,226 @@
+---
+name: gpt-researcher
+description: GPT Researcher is an autonomous deep research agent that conducts web and local research, producing detailed reports with citations. Use this skill when helping developers understand, extend, debug, or integrate with GPT Researcher - including adding features, understanding the architecture, working with the API, customizing research workflows, adding new retrievers, integrating MCP data sources, or troubleshooting research pipelines.
+---
+
+# GPT Researcher Development Skill
+
+GPT Researcher is an LLM-based autonomous agent using a planner-executor-publisher pattern with parallelized agent work for speed and reliability.
+
+## Quick Start
+
+### Basic Python Usage
+
+```python
+from gpt_researcher import GPTResearcher
+import asyncio
+
+async def main():
+    researcher = GPTResearcher(
+        query="What are the latest AI developments?",
+        report_type="research_report",  # or detailed_report, deep, outline_report
+        report_source="web",            # or local, hybrid
+    )
+    await researcher.conduct_research()
+    report = await researcher.write_report()
+    print(report)
+
+asyncio.run(main())
+```
+
+### Run Servers
+
+```bash
+# Backend
+python -m uvicorn backend.server.server:app --reload --port 8000
+
+# Frontend
+cd frontend/nextjs && npm install && npm run dev
+```
+
+---
+
+## Key File Locations
+
+| Need | Primary File | Key Classes |
+|------|--------------|-------------|
+| Main orchestrator | `gpt_researcher/agent.py` | `GPTResearcher` |
+| Research logic | `gpt_researcher/skills/researcher.py` | `ResearchConductor` |
+| Report writing | `gpt_researcher/skills/writer.py` | `ReportGenerator` |
+| All prompts | `gpt_researcher/prompts.py` | `PromptFamily` |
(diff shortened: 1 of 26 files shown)


assafelovic:
Hey @JasonBroderick this is super awesome love the feature! There are some conflicts if you can please take a look. I will dive into this soon as it requires changes to core. Thanks!

assafelovic:
@JasonBroderick sorry for late response on this this is big love it. can you please review conflicts and retest before i accept and merge?

assafelovic:
Reviewed — no longer applies to `main` after #2074.

Checkpoint persistence and retry for deep research is a genuinely useful direction; deep research is the longest, most expensive path in the system and the one where a mid-run failure costs the most.

The obstacle is shape rather than substance: **3,184 lines across 26 files**, and a large share is a `.agents/skills/gpt-researcher/` documentation tree that has nothing to do with checkpointing. Splitting the checkpoint/retry logic from the docs would turn this from a PR nobody can review into one that can actually be judged on its merits.

PR #1683, opened 2026-03-20
update exa retriever description and default search type

theishangoswami:
Updated the Exa retriever description and changed default search type from neural to auto since auto is the recommended default now.


assafelovic:
Hey @theishangoswami there are many other updates here like avian and so on non related to the PR title around exa. Can you share some more details and motivation?

PR #1700, opened 2026-03-23
Fix: resolve unhashable dict error in detailed report context deduplication

3mk4yl:
### Describe the bug
When running a detailed report, the application crashes with a `TypeError: unhashable type: 'dict'` during the subtopic generation phase. 

### To Reproduce
Run a standard research task using the Detailed Report type. The crash happens before and after subtopic generation.

### The Cause
It appears the `global_context` list now contains dictionaries instead of plain strings. The code attempts to deduplicate the context using `list(set(...))`, which crashes because dictionaries are unhashable.

### The Fix
Replace the `list(set(...))` deduplication with a dictionary-safe loop. 

**File:** `backend/report_type/detailed_report/detailed_report.py`

**Line 148:**
```python
# Change this:
# subtopic_assistant.context = list(set(self.global_context))

# To this:
unique_context = []
for item in self.global_context:
    if item not in unique_context:
        unique_context.append(item)
subtopic_assistant.context = unique_context
```

**Line 179:**

```python
# Change this:
# self.global_context = list(set(subtopic_assistant.context))

# To this:
updated_context = []
for item in subtopic_assistant.context:
    if item not in updated_context:
        updated_context.append(item)
self.global_context = updated_context
```

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,5 +1,6 @@
 OPENAI_API_KEY=
 TAVILY_API_KEY=
+XQUIK_API_KEY=
 DOC_PATH=./my-docs
 
 # NEXT_PUBLIC_GPTR_API_URL=http://0.0.0.0:8000  # Defaults to localhost:8000 if not set
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@
 
 # 🔎 GPT Researcher
 
-**GPT Researcher is an open deep research agent designed for both web and local research on any given task.** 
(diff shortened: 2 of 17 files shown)


assafelovic (approved):

assafelovic:
@3mk4yl can you please review conflict and resolve? thanks for this, would love to merge after

3mk4yl:
@assafelovic Merge conflicts are all resolved! Thanks for your help and the review.

PR #1726, opened 2026-04-04
Added missing import

varadfromeast:
Issue 1712
 This fixes the web UI failure in `backend/server/websocket_manager.py` caused by a missing `os` import, and also cleans up related backend import path issues so the server modules import correctly as a
  package.

  I added regression smoke tests for:
  - `backend.server.websocket_manager`
  - `backend.server.app`
  - `WebSocketManager.start_streaming()` reading `CONFIG_PATH`

  Verification:
  - Installed `pytest` and `pytest-asyncio` in the project venv
  - Ran `tests/test_websocket_manager.py`
  - Result: `3 passed`

  This should make similar import-path regressions much easier to catch earlier.

### DIFF
--- a/backend/server/app.py
+++ b/backend/server/app.py
@@ -29,7 +29,7 @@ from server.server_utils import (
 )
 
 from server.websocket_manager import run_agent
-from utils import write_md_to_word, write_md_to_pdf
+from backend.utils import write_md_to_word, write_md_to_pdf
 from gpt_researcher.utils.enum import Tone
 from chat.chat import ChatAgentWithMemory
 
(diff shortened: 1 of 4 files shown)


varadfromeast:
Review required @assafelovic

assafelovic:
@varadfromeast thanks for this already merged a fix to it can you verify latest and update if this is needed?

varadfromeast:
@assafelovic Raised another PR: https://github.com/assafelovic/gpt-researcher/pull/1740

The fix is good, just one small pointer -> above PR  prevents the code from crashing when run by external test tools that don't share the server's custom path settings. 

Thank you

PR #1737, opened 2026-04-16
fix: skip empty LLM streaming chunks and flush stdout

### LINKED ISSUE #1738: Streaming output silent during report generation — empty string chunks not flushed
## Bug Description

When using LLM providers that emit empty string chunks (`""`) before real content (common with thinking/reasoning models like GLM-5, DeepSeek-R1, etc.), the streaming output in `stream_response()` produces no stdout output during the report-writing phase. The report eventually completes correctly, but users see ~2-3 minutes of silence with no progress indication.

## Root Cause

In `gpt_researcher/llm_provider/generic/base.py`, the `stream_response()` function checks:

```python
if content is not None:
    paragraph += content
```

Empty string `""` passes `is not None`, so `paragraph += ""` accumulates nothing. The flush trigger (`"\n"` in content) never fires during the empty-chunk phase. When real content finally arrives, it may not contain `"\n"` either (e.g., markdown without trailing newlines), so the accumulated text is never flushed to stdout.

Additionally, `_send_output()` has the same issue — it checks `if content is not None:` but doesn't skip empty strings, leading to unnecessary WebSocket broadcasts of empty content.

## Steps to Reproduce

1. Configure GPT Researcher with a thinking/reasoning model (e.g., DeepInfra GLM-5, or any model that sends empty deltas before content)
2. Run a research query via CLI: `python -m gpt_researcher.agent "your query" --report_type research_report`
3. Observe: the "Writing report" phase produces no stdout output for ~2-3 minutes
4. The final report is correct, but the streaming experience is broken

## Expected Behavior

Streaming chunks should appear on stdout as they arrive from the LLM, with a final flush ensuring no content is lost even if the stream ends without a newline.

## Environment

- GPT Researcher: latest (`main` branch)
- Python: 3.12
- LLM: DeepInfra GLM-5 (also reproducible with other thinking models)
- OS: Linux (headless/CLI mode)

## Fix

PR #1737 addresses this with two minimal changes:
1. Skip empty strings in `stream_response()`: `if content is not None and content != "":`
2. Final flush after the stream loop to emit any remaining buffered text

kiranvk-2011:
## Summary

Fixes #1738 — Streaming output is silent during report generation when LLM providers emit empty string chunks before real content.

## Problem

Thinking/reasoning models (GLM-5, DeepSeek-R1, etc.) commonly send ~500 empty string `""` chunks before actual content arrives. In `stream_response()`, the check `if content is not None:` passes for empty strings, so `paragraph += ""` accumulates nothing and the `"\n"` flush trigger never fires. Users see ~2-3 minutes of silence during report writing.

Additionally, if the final content doesn't end with `"\n"`, the last buffered paragraph is never flushed to stdout.

## Fix

Two minimal changes in `gpt_researcher/llm_provider/generic/base.py`:

1. **Skip empty strings** in both `stream_response()` and `_send_output()`:
   `if content is not None:` → `if content is not None and content != "":`

2. **Final flush** after the stream loop in `stream_response()` to emit any remaining buffered text that didn't end with a newline.

## Testing

Verified end-to-end with DeepInfra GLM-5 — streaming output now appears correctly on stdout during report generation. No behavioral change for providers that don't emit empty chunks.

### DIFF
--- a/gpt_researcher/llm_provider/generic/base.py
+++ b/gpt_researcher/llm_provider/generic/base.py
@@ -294,12 +294,13 @@ class GenericLLMProvider:
         # Streaming the response using the chain astream method from langchain
         async for chunk in self.llm.astream(messages, **kwargs):
             content = chunk.content
-            if content is not None:
-                response += content
-                paragraph += content
-                if "\n" in paragraph:
-                    await self._send_output(paragraph, websocket)
-                    paragraph = ""
+            if not content:
+                continue
+            response += content
+            paragraph += content
+            if "\n" in paragraph:
+                await self._send_output(paragraph, websocket)
+                paragraph = ""
 
         if paragraph:
             await self._send_output(paragraph, websocket)
@@ -310,7 +311,7 @@ class GenericLLMProvider:
         if websocket is not None:
             await websocket.send_json({"type": "report", "output": content})
         elif self.verbose:
-            print(f"{Fore.GREEN}{content}{Style.RESET_ALL}")
+            print(f"{Fore.GREEN}{content}{Style.RESET_ALL}", flush=True)
 
 
 def _check_pkg(pkg: str) -> None:


assafelovic (approved):
this is great thank you!

kiranvk-2011:
> this is great thank you!

@assafelovic thank you for reviewing and incorporating my contribution.

PR #1756, opened 2026-04-28
fix: limit concurrent FireCrawl API calls to prevent rate limit failures

### LINKED ISSUE #1602: FireCrawl scraper returns empty content under concurrent requests in deep research mode
## Bug Description

When using `SCRAPER=firecrawl` in deep research mode, **~71% of URLs return empty content** ("Content too short or empty"), even though the same URLs succeed when scraped individually. The issue is caused by Firecrawl's API rate limits being exceeded during concurrent scraping.

## Steps to Reproduce

1. Set `SCRAPER=firecrawl` with a Firecrawl Free Tier API key
2. Configure deep research: `DEEP_RESEARCH_BREADTH=3`, `DEEP_RESEARCH_DEPTH=2`, `DEEP_RESEARCH_CONCURRENCY=4`
3. Run a deep research query
4. Observe ~70%+ of URLs failing with "Content too short or empty"

### Deep Research Test (concurrent)
| Metric | Value |
|--------|-------|
| Total scraper calls | 156 (FireCrawl: 137, PyMuPDF: 19) |
| Successful | 45 |
| Failed (content too short) | **111** |
| **Success rate** | **28.8%** |

### Individual Test (sequential, same URLs)
All 6 tested URLs that failed during deep research **succeeded** when tested individually:

| URL | Deep Research | Individual Test |
|-----|-------------|----------------|
| foodnext.net | ❌ empty | ✅ 10,795 chars |
| tw.news.yahoo.com | ❌ empty | ✅ 75,509 chars |
| brain.com.tw | ❌ empty | ✅ 23,984 chars |
| businessweekly.com.tw | ❌ empty | ✅ 14,646 chars |
| fc.bnext.com.tw | ❌ empty | ✅ 11,918 chars |
| csrone.com | ❌ empty | ✅ 4,892 chars |

## Root Cause

Firecrawl Free Tier limits:
- **10 requests/minute** for `/scrape`
- **2 concurrent browsers**

GPT Researcher's deep research mode sends many requests in parallel (`asyncio.gather` in `scraper.py` line 50-51), easily exceeding these limits. When rate limited, Firecrawl appears to return empty/minimal content rather than an explicit error.

```python
# scraper.py - all URLs scraped concurrently
)
```

### Option A: Add rate limiting to the Scraper class

```python
# Add a semaphore or rate limiter for API-based scrapers
import asyncio

class Scraper:
    def __init__(self, urls, user_agent, scraper, worker_pool):
        ...
        # Rate limit for API-based scrapers
        if self.scraper in ("firecrawl", "tavily_extract"):
            self._api_semaphore = asyncio.Semaphore(2)  # max 2 concurrent
        
    async def extract_data_from_url(self, link, session):
        if hasattr(self, '_api_semaphore'):
            async with self._api_semaphore:
                return await self._extract_data(link, session)
        return await self._extract_data(link, session)
```

### Option B: Add retry with backoff for empty results

If the scraper returns empty content for a URL that should have content, retry with exponential backoff.

### Option C: Document rate limit configuration

Allow users to configure concurrency limits per scraper type:

```python
os.environ["FIRECRAWL_CONCURRENCY"] = "2"
os.environ["FIRECRAWL_REQUESTS_PER_MINUTE"] = "10"
```

## Environment

- gpt-researcher: 0.14.5
- Firecrawl: Free Tier (10 req/min, 2 concurrent browsers)
- Python: 3.13.5
- Deep Research config: breadth=3, depth=2, concurrency=4

octo-patch:
Fixes #1602

## Problem

When using `SCRAPER=firecrawl` in deep research mode, GPT Researcher launches many parallel URL scrapes via `asyncio.gather`. FireCrawl's Free Tier allows only **2 concurrent browsers**, so most requests silently return empty content instead of an explicit error — resulting in ~71% failure rates observed by users.

## Solution

Add a module-level `asyncio.Semaphore` shared across all `FireCrawl` instances, and expose a `scrape_async()` method that acquires the semaphore before each API call. The existing `scraper.py` already prefers `scrape_async()` over `scrape()` when the method exists (line 121–122), so no changes outside `firecrawl.py` are needed.

The concurrency limit defaults to `2` (matching the Free Tier browser limit) and is configurable via `FIRECRAWL_CONCURRENCY`:

```bash
# Free Tier (default)
FIRECRAWL_CONCURRENCY=2

# Hobby/Paid tiers with higher concurrency
FIRECRAWL_CONCURRENCY=10
```

## Testing

The fix is a targeted, additive change: `scrape()` is unchanged, and `scrape_async()` simply wraps it with a semaphore before running it in an executor. Existing behaviour is preserved for all non-FireCrawl scrapers.


assafelovic:
Landed in `main` via #2074 — commit `7f8313b2` (*fix: cap concurrent FireCrawl calls*).

**Note:** only the FireCrawl and config hunks were taken. The branch had also picked up unrelated docs/blog and image-generator changes that conflict with current `main`. It also shipped without tests, so concurrency tests were written for it — they assert the observed overlap never exceeds the cap, at the default and under an env override.

PR #1775, opened 2026-05-16
fix(llm): raise max_tokens cap from 32k to 200k for modern long-output models

### LINKED ISSUE #1335: Deep Research: Reports are cut off
**Describe the bug**
Research output is cut off when using method "deep"

**To Reproduce**
1. Using report type deep and models from Anthropic (3.5, 3.7)
2. Generate report
3. Get a report which is cut off at the end

**Expected behavior**
A full report which is not cut off and has a list of sources at the end.

**Screenshots**

Image

RomainROCH:
## Problem

`create_chat_completion` in `gpt_researcher/utils/llm.py` rejects any `max_tokens > 32001` with a hardcoded `ValueError`. This predates modern LLMs with 64k–128k output capacity (Claude 4.x family, GPT-5 family) and silently breaks long-form report generation:

1. The user sets `SMART_TOKEN_LIMIT=64000` to leverage their model
2. Report generation hits the cap and raises `ValueError`
3. `report_generation.py` catches it in a broad `except` and returns `""`
4. The user gets an empty report with no actionable error

This bug was reported in #1335 (April 2025), closed without a verified fix. The cap is still present in `main` as of today.

## Fix

The cap is **raised, not removed**. It still serves as a sanity guard against absurd typos (e.g., `SMART_TOKEN_LIMIT=1000000`), which matters for providers like Ollama or litellm-routed backends where the upstream API may not surface a clear error. Cloud APIs (Anthropic, OpenAI) already enforce model-specific limits with descriptive errors, so we defer to them for the precise per-model boundary.

The new cap is 200k, leaving ~50% headroom above the largest currently available cloud output limit (128k). The error message now points at the relevant env vars instead of stating an arbitrary number.

Three atomic commits:

1. **fix(llm)**: raise the cap from 32k to 200k, improve error message, add unit tests covering both accepted (64k, 128k) and rejected (1M) cases.

2. **docs(env)**: add commented `TOKEN_LIMIT` examples to `.env.example` with recommended values per model class. Default behavior unchanged.

3. **docs(config)**: correct outdated default values in `config.md` (runtime defaults are 3000/6000/4000, docs said 2000/4000/4000) and add a recommended-values table for modern models.

## Related follow-ups (out of scope)

- `report_generation.py:290-306` swallows exceptions silently and returns `""`. The 32k cap was its main trigger, but other failures will still surface as empty reports. Worth a separate PR with proper error propagation.
- `agent_creator.py` and `tools.py` still use hardcoded `max_tokens=4000` defaults independent of config. Same pattern, separate sweep.

## Tests

- `tests/test_llm_max_tokens.py` (new): asserts that `create_chat_completion` accepts `max_tokens` values previously blocked (64000, 128000), and still rejects absurd values (1000000) with an actionable error message.
- New test passes locally in asyncio strict mode.

Closes #1335

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -44,3 +44,21 @@ DOC_PATH=./my-docs
 # LANGCHAIN_ENDPOINT="https://api.smith.langchain.com"
 # LANGCHAIN_API_KEY=
 # LANGCHAIN_PROJECT="gpt-researcher"
+
+# Token Limits
+# ------------
+# Maximum output tokens per LLM call. Leaving unset uses package defaults
+# (FAST=3000, SMART=6000, STRATEGIC=4000) suited to GPT-4o-class models.
+# Modern long-output models (Claude 4.x, GPT-5) truncate reports at
+# these defaults.
+#
+# Recommended values by model class:
+#   GPT-4o family (16k max output):   SMART_TOKEN_LIMIT=8000
+#   Claude Haiku 4.5 (64k max):       SMART_TOKEN_LIMIT=16000
+#   Claude Sonnet 4.6 (64k max):      SMART_TOKEN_LIMIT=16000
+#   Claude Opus 4.7 (128k max):       SMART_TOKEN_LIMIT=32000
+#   GPT-5 family (128k max):          SMART_TOKEN_LIMIT=32000
+#
+# FAST_TOKEN_LIMIT=8000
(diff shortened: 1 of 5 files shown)


assafelovic (approved):
That's a great point thank you

PR #1779, opened 2026-05-19
Fix incorrect config file path in Chinese README

wesssI:
This PR fixes incorrect configuration file references in the Chinese README.

### What was wrong
The README pointed users to `config/config.py`, which is not the correct file for updating default model/search settings.

### What I changed
- Updated the referenced path to `gpt_researcher/config/variables/default.py` in the Chinese README.
- Corrected related wording so users can find the right file directly.

### Why this matters
New contributors and users can now update LLM/search defaults without confusion, reducing setup friction.

### Scope
- Documentation only
- No runtime/code behavior changes


assafelovic (approved):
Thank you would love your help continuing with this!

Issue #1685, opened 2026-03-21
Feature request: anybrowse MCP integration for Cloudflare-protected scraping


kc23go:
## Problem

GPT-Researcher's web retrieval fails silently on Cloudflare-protected sites (major news outlets, LinkedIn, Amazon, government sites). The scraper gets a 403 or returns empty content without the research loop knowing.

## Proposed integration

[anybrowse](https://anybrowse.dev) is an MCP-native web scraper with Cloudflare bypass via residential Chrome. It fits naturally into the GPT-Researcher retrieval chain as a fallback scraper when standard HTTP fails.

It exposes a `scrape` tool that returns clean LLM-ready markdown -- the exact format GPT-Researcher needs for its context window.

## Quick test

Add as an MCP server:

```json
{
  "mcpServers": {
    "anybrowse": {
      "type": "streamable-http",
      "url": "https://anybrowse.dev/mcp"
    }
  }
}
```

Then: "Research the latest AI regulation news from EU government sites" -- anybrowse handles the Cloudflare-protected `.europa.eu` domains that standard scrapers bounce off.

## Details

- Free tier: 10 scrapes/day, no API key
- Paid: $5 for 3,000 credits (never expire)
- MCP tools: `scrape`, `crawl`, `batch_scrape`, `extract`, `search`
- Docs: https://anybrowse.dev/docs

Would be happy to contribute a retriever class if this seems like a good fit for the project.

assafelovic:
Closing as a duplicate of #1684 — let's keep the discussion consolidated there. Thanks!

Issue #1404, opened 2025-05-20
BUG: Self hosted firecrawl returning 0 scraped results


ga-it:
**Describe the bug**
When using a self-hosted firecrawl for scraping, it always returns 0 scraped results.

When calling firecrawl.scrape_url(), the returned response object does not behave like a standard dict (e.g., response["metadata"]["statusCode"], response.data.markdown). Attempting to access these keys directly results in attribute or key errors.

Instead, the response appears require access via dot notation (e.g., response.metadata, response.markdown). This inconsistency causes downstream failures.

**To Reproduce**
1. For web scraping, use a self-hosted firecrawl
2. Start a new research query
3. Wait for scraping results to appear
4. Observe 0 results scraped

**Expected behavior**
Of the 10 pages scraped, some should have returned results.

**Fix**
The following changes resolved the issue in my case. In app/gpt_researcher/scraper/firecrawl/firecrawl.py, inside the scrape method's try block:

I modified the following three lines:

- elif response["metadata"]["statusCode"] != 200:  ->  elif response.metadata["statusCode"] != 200:

- content = response.data.markdown ->  content = response.markdown

- title = response["metadata"]["title"] ->   title = response.metadata["title"]

anzi-tech:
Thanks for this fix! Can confirm this is a working solution to the problem.

assafelovic:
Closing as part of a long-tail backlog cleanup — this issue has been open for **over a year** with no recent activity. This is not a judgment on its validity; we're clearing the aged backlog so the team can focus on current, reproducible issues.

**If this is still relevant on the latest version, please reopen (or comment) with up-to-date details — version, config, and a repro — and we'll gladly take another look.** Thank you for the report! 🙏

Issue #1081, opened 2025-01-17
Cloudflare Protection Blocking Crunchbase Results


PatricioCabo:
**Is your feature request related to a problem? Please describe.**
When trying to scrape Crunchbase pages through Tavily, the results are being blocked by Cloudflare's security protection. This affects the quality of company research as Crunchbase is a crucial source for company information.

#### Current Behavior
- Tavily attempts to scrape Crunchbase URLs (e.g., "https://www.crunchbase.com/organization/XXXXX")
- Instead of getting company data, we receive Cloudflare's block page:
  ```
  "please enable cookies. sorry, you have been blocked you are unable to access crunchbase.com..."
  ```
- This causes our similarity check to fail as the content doesn't contain company information

**Describe the solution you'd like**
- Tavily should be able to bypass Cloudflare protection
- Successfully retrieve company information from Crunchbase pages

**Describe alternatives you've considered**
- Implement headless browser support in Tavily (e.g., Playwright, Puppeteer)
- Add proper headers and cookie management

ElishaKay:
Welcome @PatricioCabo 

Agreed.
It would be great if someone from the community would take a crack at adding [Tavily Extract](https://docs.tavily.com/docs/python-sdk/tavily-extract/getting-started) functionality to GPTR.

Green light for whoever feels the calling

assafelovic:
Closing as part of a long-tail backlog cleanup — this issue has been open for **over a year** with no recent activity. This is not a judgment on its validity; we're clearing the aged backlog so the team can focus on current, reproducible issues.

**If this is still relevant on the latest version, please reopen (or comment) with up-to-date details — version, config, and a repro — and we'll gladly take another look.** Thank you for the report! 🙏

PR #1786, opened 2026-05-27
docs(proposal): Firefox-based stealth scraper backend

feder-cr:
opening as draft to check interest before building this out further.

would an optional `invisible_firefox` scraper backend fit in scope, parallel to firecrawl / browser / web_base_loader / tavily_extract under `gpt_researcher/scraper/`?

motivation: research-relevant pages behind Cloudflare/Akamai/Datadome/hCaptcha currently return empty content or 403. relevant open issues: #1685, #1081, #1602, #1404.

the backend wraps invisible_playwright (github.com/feder-cr/invisible_playwright, also on PyPI as invisible-playwright), which drives a patched Firefox through the normal Playwright context/page API. the C++ level patches live in a separate repo, firefox_antidetect_patch. selected via `SCRAPER=invisible_firefox`, optional dependency, no change to defaults.

two corrections to the original post. first, this diff already includes the actual scraper class, not just a docs stub - I described it wrong the first time. second, the module/class name and the install line in the docstring still reference invisible_firefox, a separate profile-manager project of mine that's since been retired; the code here never depended on it, it only imports invisible_playwright, so before this leaves draft I'll rename the module/class and switch the install line to `pip install invisible-playwright`.

if the answer is "not in scope" I'll close it without noise.

### DIFF
--- a/gpt_researcher/scraper/__init__.py
+++ b/gpt_researcher/scraper/__init__.py
@@ -6,6 +6,7 @@ from .browser.browser import BrowserScraper
 from .browser.nodriver_scraper import NoDriverScraper
 from .tavily_extract.tavily_extract import TavilyExtract
 from .firecrawl.firecrawl import FireCrawl
+from .invisible_firefox.invisible_firefox_scraper import InvisibleFirefoxScraper
 from .scraper import Scraper
 
 __all__ = [
@@ -18,4 +19,5 @@ __all__ = [
     "TavilyExtract",
     "Scraper",
     "FireCrawl",
+    "InvisibleFirefoxScraper",
 ]
--- a/gpt_researcher/scraper/__init__.py
+++ b/gpt_researcher/scraper/__init__.py
@@ -6,7 +6,7 @@ from .browser.browser import BrowserScraper
 from .browser.nodriver_scraper import NoDriverScraper
 from .tavily_extract.tavily_extract import TavilyExtract
 from .firecrawl.firecrawl import FireCrawl
-from .invisible_firefox.invisible_firefox_scraper import InvisibleFirefoxScraper
+from .invisible_playwright.invisible_playwright_scraper import InvisiblePlaywrightScraper
 from .scraper import Scraper
 
 __all__ = [
@@ -19,5 +19,5 @@ __all__ = [
     "TavilyExtract",
     "Scraper",
     "FireCrawl",
-    "InvisibleFirefoxScraper",
+    "InvisiblePlaywrightScraper",
 ]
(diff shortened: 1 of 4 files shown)


feder-cr:
@assafelovic any interest in an optional stealth scraper backend for sources behind anti-bot? it's opt-in, no change to defaults. happy to close otherwise.

assafelovic:
hey @feder-cr this requires 49 files of changes. is that necessary? this is a heavy change i am concerned pushing for the business impact it brings. wdyt?

feder-cr:
totally fair, 49 files is too much for an optional backend, and most of that is tests and docs. i can cut it to the minimal version: one scraper module behind an env flag, defaults untouched, no new required deps. that's 3-4 files. if the slimmed-down shape works better i'll push it, otherwise no hard feelings closing this.

feder-cr:
Slimmed it down as promised - it is now just the opt-in scraper (4 files, +62): one InvisibleFirefoxScraper that renders the page with invisible_playwright then parses it with the existing bs4 utils, registered under SCRAPER=invisible_firefox. Defaults untouched, and invisible_playwright is a lazy import so it is not a required dependency. Dropped everything unrelated. Tested end to end (renders + extracts content/title/images on real pages). Let me know if the shape works and I can add a short docs note.

feder-cr:
@assafelovic slimmed this down to the minimal version as discussed, just checking if it's closer to what you had in mind now.

feder-cr:
updated the description above. two things were wrong in it: I attributed the C++-level Firefox patches to invisible_firefox when that's actually a separate repo (firefox_antidetect_patch) - and invisible_firefox itself, a profile-manager project of mine, has since been retired. the code in this diff never depended on it though, it only imports invisible_playwright, so I fixed the text instead of pulling the PR. also corrected the "docs stub" line, this already ships the working scraper class, not just a proposal doc. will rename the module/class before taking this out of draft.

assafelovic:
Reviewed — this is still marked **draft**, and reads as a proposal rather than a finished backend, so I'm not reviewing it as though it were ready.

On the substance: a stealth/anti-bot scraper backend addresses something real. #2074 landed detection for anti-bot challenge pages (they were previously ingested as article content and polluted reports), and #1684 is an open request for Cloudflare-protected scraping — so the need is demonstrated, not hypothetical.

Two things worth settling in the proposal before code: how a Firefox/Playwright dependency stays optional rather than becoming a required install, and where the project wants to sit on scraping sites that are actively blocking automation.

**Also: this targets `master`, which is no longer the trunk.** `main` is the default branch and is now 160 commits ahead; `master` last moved in April. Retargeting to `main` is needed regardless of anything else.

CI now runs on every PR — `imports` and `unit` across Python 3.11–3.14, plus `collect` — so once it's rebased you'll get a direct answer.

PR #1799, opened 2026-06-07
Abstain instead of fabricating a report when no content was gathered

aaravanmay:
While testing how GPT Researcher handles empty retrievals, I found it produces a confident, fully-sourced-looking report (with invented citations) when *every* source returns no content — searches rate-limited, blocked, or genuinely empty. `ReportGenerator.write_report` passes the (empty) context straight to the report LLM with no check, so the model fills in from its priors.

This adds a guard: if no research content was gathered, abstain with a clear message instead of writing. Includes a regression test (mocks the report LLM, so no API call) that fails on `main` and passes with the change. Happy to make it a config flag if you'd rather keep the current behavior opt-in.

### DIFF
--- a/gpt_researcher/skills/writer.py
+++ b/gpt_researcher/skills/writer.py
@@ -75,6 +75,17 @@ class ReportGenerator:
             )
 
         context = ext_context or self.researcher.context
+
+        # Guard against fabricating a report from nothing: if no research content was
+        # gathered (every retriever returned empty / was blocked / rate-limited), don't
+        # silently write a confident, sourced-looking report - abstain so it is visible.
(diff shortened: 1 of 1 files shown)


aaravanmay:
The fabricate-on-empty-retrieval bug in this PR is one instance of a class: the agent acting confidently on empty/stale/truncated tool data. I built faultline (open source) to catch that class deterministically in CI — the abstain-on-empty invariant is ~5 lines. I can PR a suite + a shadow-mode GitHub Action (continue-on-error, never blocks merges) so you just see what it flags for a week. May I?

assafelovic (approved):
you're right, this is long overdue!

PR #1802, opened 2026-06-09
fix(deep-research): keep oversized context from collapsing to empty

### LINKED ISSUE #1770: Deep mode: writer receives empty {context} placeholder; report rejection across all params (post-PR-#1637, current master)
## Summary

After upgrading from `0.12.8` to current master (post-PR-#1637), `report_type="deep"` consistently completes the deep research phase but produces an **empty `{context}` placeholder** in the writer prompt template. The writer LLM correctly refuses to fabricate, returning a "no research material provided" rejection rather than a synthesized report.

This is a **different symptom from issue #1378** (which describes truncated/incomplete output). Here the writer never starts: the prompt arrives with `""` (empty string) where the research context should be.

## Reproduction

Verified across multiple param combinations, all rejecting:

| breadth | depth | reasoning_effort | total_words | result |
|---|---|---|---|---|
| 2 | 2 | medium | 400 | rejection |
| 4 | 2 | high | 2000 | rejection |
| 2 | 3 | high | 2500 | rejection |
| 4 | 2 | medium | 2000 | rejection |

Same query and config produce a valid report when `report_type="detailed_report"` is used instead. Detailed mode is fully functional in current master.

Same Docker image, same `.env`, same SearXNG retriever. Bug appears only when `report_type="deep"`.

## Verbatim writer prompt arriving empty

Rejection messages from the writer reference the template literally:

> "Using the following hierarchically researched information and citations: `\`\``"

The backticks are literal quotation marks around what should be the populated `{context}`. The placeholder in [`prompts.py:458`](https://github.com/assafelovic/gpt-researcher/blob/master/gpt_researcher/prompts.py#L458) is:

```python
return f"""
Using the following hierarchically researched information and citations:

"{context}"

Write a comprehensive research report answering the query: "{question}"
...
"""
```

So `{context}` is being substituted with `""` (empty string) rather than the deep research output.

## Pipeline trace

1. `agent.conduct_research()` routes to `_handle_deep_research()` because `self.report_type == ReportType.DeepResearch.value`.
2. `_handle_deep_research()` calls `self.context = await self.deep_researcher.run(on_progress)`.
3. `deep_research.DeepResearchSkill.run()` aggregates `context_with_citations` from learnings + results, then calls:
   ```python
   final_context = trim_context_to_word_limit(context_with_citations)
   self.researcher.context = "\n".join(final_context)
   return self.researcher.context
   ```
4. `agent.write_report()` then calls `self.report_generator.write_report(...)`, which does `context = ext_context or self.researcher.context`.
5. `generate_report` calls `generate_deep_research_prompt(question, context, ...)`.
6. Prompt template uses `f'"{context}"'` → `""` if context is empty string.

The empty string therefore implies `final_context` is `[]` (or `"\n".join([])`).

## Likely cause (un-proven but suspected)

`trim_context_to_word_limit` at [`deep_research.py:23-37`](https://github.com/assafelovic/gpt-researcher/blob/master/gpt_researcher/skills/deep_research.py#L23) breaks out of the loop and returns `[]` if the **first item** in `reversed(context_list)` already exceeds `MAX_CONTEXT_WORDS` (25000):

```python
for item in reversed(context_list):
    words = count_words(item)
    if total_words + words <= max_words:
        trimmed_context.insert(0, item)
        total_words += words
    else:
        break    # ← if first item > 25000 words, returns []
```

When PR #1637 fixed `count_words` to handle `list` input, the function began counting items that produce very high word counts (e.g., serialized lists of source-dict chunks). The trim then silently empties the context.

In `0.12.x` the `count_words` crash on `list` inputs masked this latent trim behavior. Now that `count_words` survives list inputs, the trim path runs and aggressively zeros out the context.

I have not yet confirmed via verbose logs whether `len(all_context) → 0` at the trim log line:
```
logger.info(f"Trimmed context from {len(all_context)} items to {len(trimmed_context)} items to stay within word limit")
```

If a maintainer can enable that log on a repro, confirmation should be immediate.

## Suggested fix (defensive)

Always include at least one item in the trimmed context, truncating tokens if needed:

```python
def trim_context_to_word_limit(context_list, max_words=MAX_CONTEXT_WORDS):
    total_words = 0
    trimmed_context = []
    for item in reversed(context_list):
        words = count_words(item)
        if total_words + words <= max_words:
            trimmed_context.insert(0, item)
            total_words += words
        elif not trimmed_context:
            # Always include at least one item, truncated to max_words tokens
            text = str(item) if not isinstance(item, str) else item
            tokens = text.split()[:max_words]
            trimmed_context.insert(0, " ".join(tokens))
            break
        else:
            break
    return trimmed_context
```

Alternatively, treat `MAX_CONTEXT_WORDS` as an env-configurable parameter, since 25k is a hard cap that may be hostile to deep+high-reasoning workloads on modern long-context models.

## Environment

- `gpt-researcher` from master @ commit ~`92bfc038` (Apr 16, 2026)
- Python 3.12, image built from the project's `Dockerfile`
- Retriever: SearXNG (self-hosted)
- LLM: `openai:gpt-5.x` via OpenAI-compatible proxy
- Embeddings: Ollama `qwen3-embedding:0.6b`
- `MAX_ITERATIONS=4`, `MAX_SEARCH_RESULTS_PER_QUERY=8`, `CURATE_SOURCES=True`, `REASONING_EFFORT={medium|high}`

## Workaround in place

Falling back to `report_type="detailed_report"` for all real research. Detailed mode produces excellent output with the same query and params; no rejection.

## Related

- #1378 — `write_report()` yields incomplete/inconsistent results (deep-only) — same problem class but different symptom (truncated, not empty rejection)
- PR #1637 — fixed `count_words` crash, fixed in master Feb 2026, likely surfaced this latent trim behavior
- #1332, #1279 — original `count_words` crash bugs (resolved by #1637)

XiaojuCH:
Fixes #1770.

This PR prevents deep-research context trimming from collapsing to an empty string when the first retained context item alone exceeds `MAX_CONTEXT_WORDS`.

What changed:
- update `trim_context_to_word_limit()` in `gpt_researcher/skills/deep_research.py`
- when the first candidate item already exceeds the cap and nothing has been retained yet, keep a truncated version of that item instead of returning an empty context
- add focused regression tests in `tests/test_deep_research_parsing.py`

Why:
- deep research currently joins `final_context` into `self.researcher.context`
- if `trim_context_to_word_limit()` returns `[]`, the writer receives `""` for `{context}`
- this matches the failure mode described in #1770, where deep research completes but the writer rejects the report because no research material is present

Validation:
- `.\.venv\Scripts\python.exe -m pytest tests/test_deep_research_parsing.py -q`
- `25 passed`

### DIFF
--- a/gpt_researcher/skills/deep_research.py
+++ b/gpt_researcher/skills/deep_research.py
@@ -217,10 +217,14 @@ def trim_context_to_word_limit(context_list: List[str], max_words: int = MAX_CON
 
     # Process in reverse to keep most recent items
     for item in reversed(context_list):
+        text = " ".join(str(part) for part in item) if isinstance(item, list) else str(item)
         words = count_words(item)
         if total_words + words <= max_words:
             trimmed_context.insert(0, item)  # Insert at start to maintain original order
             total_words += words
+        elif not trimmed_context:
+            trimmed_context.insert(0, " ".join(text.split()[:max_words]))
+            break
         else:
             break
 
--- a/tests/test_deep_research_parsing.py
+++ b/tests/test_deep_research_parsing.py
@@ -5,9 +5,11 @@ import pytest
 import gpt_researcher.skills.deep_research as deep_research_module
 from gpt_researcher.skills.deep_research import (
     DeepResearchSkill,
+    MAX_CONTEXT_WORDS,
     parse_follow_up_questions_response,
     parse_research_results_response,
     parse_search_queries_response,
+    trim_context_to_word_limit,
 )
 
 
@@ -294,3 +296,21 @@ def test_parse_research_results_response_extracts_inline_legacy_url():
 
     assert result["learnings"] == ["stuff happened at"]
     assert result["citations"]["stuff happened at"] == "https://example.com/api/v1?key=value"
+
+
+def test_trim_context_to_word_limit_keeps_first_item_when_it_alone_exceeds_cap():
+    oversized_item = "word " * (MAX_CONTEXT_WORDS + 500)
+
+    result = trim_context_to_word_limit([oversized_item], max_words=MAX_CONTEXT_WORDS)
+
+    assert result, "trim_context_to_word_limit() should not return an empty context when the first item is oversized"
+    assert len(result[0].split()) == MAX_CONTEXT_WORDS
+
+
+def test_trim_context_to_word_limit_preserves_recent_context_and_keeps_one_oversized_tail_item():
+    earlier = "early context " * 50
+    oversized_latest = "latest " * (MAX_CONTEXT_WORDS + 250)
+
+    result = trim_context_to_word_limit([earlier, oversized_latest], max_words=MAX_CONTEXT_WORDS)
+
+    assert result == [" ".join(oversized_latest.split()[:MAX_CONTEXT_WORDS])]


assafelovic (approved):
this is great thank you

PR #1804, opened 2026-06-11
added image search capabilities for Royalty Free Photos and videos using Pexels, Pixabay, and Unsplash API's

sdisaacson:
Pretty much follows along with all the other image generation  capabilities, but here I"m searching for photos.


assafelovic:
@sdisaacson thank you for this. the issue i have with is it that search like pexels etc show generic images. the idea here was to show related images to the research report. wdyt? do you think i am wrong?

PR #1814, opened 2026-06-21
Add Keenable web search retriever

ilya-bogin-keenable:
Adds Keenable as a search retriever, selectable with `RETRIEVER=keenable`.

The main difference from the existing retrievers is that it doesn't require
an API key. By default it hits Keenable's keyless public endpoint, so it works
out of the box. Setting `KEENABLE_API_KEY` switches to the authenticated
endpoint and lifts the rate limits.

### Usage

```bash
RETRIEVER=keenable
```

Optional:

```bash
KEENABLE_API_KEY=...                 # lifts rate limits, enables realtime mode
KEENABLE_SEARCH_MODE=pro             # pro (default) or realtime
KEENABLE_API_URL=https://api.keenable.ai   # override base URL (HTTPS)
```

### Changes

- `gpt_researcher/retrievers/keenable/` — `KeenableSearch`, follows the Tavily
  retriever (`query` + `query_domains`, `.search()` returns `{title, href, body}`).
- Registered in `retrievers/__init__.py`, `actions/retriever.py`, and
  `VALID_RETRIEVERS`.
- Docs: added the retriever to the search engines page with a short setup section.

`query_domains` maps to Keenable's single-domain `site` filter (first domain
used). The base URL is read from the environment with HTTPS enforced; there's no
URL fetch path here, so nothing user/agent-controlled reaches the network beyond
the query.

Tested against the public endpoint and with a mocked transport (keyed/keyless
selection, headers, site filter, response mapping, error handling).

### DIFF
--- a/docs/docs/gpt-researcher/search-engines/search-engines.md
+++ b/docs/docs/gpt-researcher/search-engines/search-engines.md
@@ -99,7 +99,7 @@ so you only need to set the retriever:
 RETRIEVER=keenable
 ```
 
-To lift rate limits and enable `realtime` mode, set an API key (create one at
+To lift rate limits, set an API key (create one at
 [keenable.ai/console](https://keenable.ai/console)):
 
 ```bash
@@ -110,7 +110,6 @@ KEENABLE_API_KEY=your_api_key_here   # optional
 **Optional Configuration:**
 
 ```bash
-KEENABLE_SEARCH_MODE=pro             # "pro" (default, deeper) or "realtime" (low latency, requires a key)
 KEENABLE_API_URL=https://api.keenable.ai   # override the API base URL (HTTPS); e.g. for staging
 ```
 
--- a/docs/docs/gpt-researcher/search-engines/search-engines.md
+++ b/docs/docs/gpt-researcher/search-engines/search-engines.md
@@ -24,6 +24,7 @@ RETRIEVER=tavily, arxiv
 Thanks to our community, we have integrated the following web search engines:
 
 - [Tavily](https://app.tavily.com) - Default
+- [Keenable](https://keenable.ai) - Env: `RETRIEVER=keenable` - keyless by default - [Setup Guide](#keenable)
 - [Bing](https://www.microsoft.com/en-us/bing/apis/bing-web-search-api) - Env: `RETRIEVER=bing`
 - [Brave Search](https://brave.com/search/api/) - Env: `RETRIEVER=brave` and `BRAVE_API_KEY`
 - [Google](https://developers.google.com/custom-search/v1/overview) - Env: `RETRIEVER=google`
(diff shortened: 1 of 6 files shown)


assafelovic:
hi @ilya-bogin-keenable can you resolve the conflicts?

ilya-bogin-keenable:
Hi @assafelovic. I have resolved the conflicts

ilya-bogin-keenable:
@assafelovic the conflicts were resolved on June 29 and the PR is still mergeable, so this one is just waiting on a look.

Recap: the retriever mirrors the existing ones (`RETRIEVER=keenable`) and needs no API key, it runs against the public endpoint by default; an optional `KEENABLE_API_KEY` only lifts the rate limit. Happy to merge `master` in again if you'd like it freshened before review.

ilya-bogin-keenable:
@assafelovic checking in once more, then I will stop bumping this thread.

Since your June 23 note the PR has needed nothing: conflicts resolved June 29, and it has stayed mergeable against `master` for seven weeks without a single rebase. The diff is **+152/-0** across 6 files, so nothing existing is modified or removed. It is a new `keenable` retriever that mirrors the shape of the ones already there (`RETRIEVER=keenable`), plus its registration and a docs row.

The one thing that makes it different from the neighbours: it works with **no API key**, hitting the public endpoint by default. `KEENABLE_API_KEY` is optional and only lifts the rate limit.

Whatever the answer is, it is useful to know: is this something you want merged, is there a change you would like first, or is it not a direction you want in core? Happy to freshen `master` into it or to trim the docs part if that is what is holding it.

ilya-bogin-keenable:
Correction to my message yesterday, and a reason not to merge the previous state.

I found a real bug in the mapping while auditing our integrations, and it is now fixed on this branch (`98ae15f`). The retriever read each result's `body` from the API's `description` field. Keenable returns **both** `description` and `snippet`, and `description` is frequently empty while `snippet` carries the page text. Verified against the live keyless endpoint just now: **10/10 results came back with `len(description) == 0` and `len(snippet)` between 2000 and 2600**. So as written, the retriever handed the researcher ten results with a title, a URL and no text at all.

It now reads `snippet` first with a `description` fallback. Two details worth flagging for review:

- Snippets are raw page text, newlines included, and run an order of magnitude longer than what the other retrievers return, so the text is whitespace-collapsed and capped at 500 characters. That keeps `body` snippet-sized, which matches the assumption in `_search_relevant_source_urls` — only `raw_content` marks a result as already fetched, so each `href` is still scraped for the full page.
- Added `tests/test_keenable_retriever.py`, mirroring `test_brave_retriever.py`. It covers the mapping with a realistic fixture (empty `description`, populated `snippet`), the fallback, the whitespace collapse and the cap, url-less rows, and the keyless default. The bug was invisible to a mocked test whose fixture populates `description`, which no real response does, so the fixture is the point.

Now 7 files / +295, still additive, no existing behaviour modified. Sorry for asking you to merge the previous state.

assafelovic:
Reviewed — the patch itself is in good shape (tests, docs, error handling), but it **no longer applies to `main`**: the retriever registry and `retrievers/utils.py` moved in #2074, which landed a large batch of retriever hardening. A rebase is needed before this can be evaluated on a green CI run.

CI now runs on every PR (`imports` across 3.11–3.14, `collect`, `unit`), so a rebase will tell you directly.

---

ilya-bogin-keenable:
@assafelovic I have rebased the PR

PR #1816, opened 2026-06-22
Update web_scraping.py

Joly0:
Added code to close open requests.Session so its underlying connection pool (and the sockets it keeps alive) is released. A new Scraper/Session is created on every scrape batch; without this, the file descriptors accumulate over the lifetime of a long-running process (e.g. the gptr-mcp server) and eventually exhaust the open-file limit ([Errno 24] Too many open files), which breaks the server's ability to accept new connections.


assafelovic (approved):
great point thank you!

PR #1818, opened 2026-06-23
Add SSRF and local-file-read protection for scraped URLs

### LINKED ISSUE #1794: [Security] Unauthenticated Server-Side Request Forgery (SSRF)
Hi. Unsure if this is intended. If ssrf isnt, I’ll like to report the details privately.

### LINKED ISSUE #1805: Unauthenticated arbitrary local PDF file read via source_urls and the PyMuPDFScraper local-file branch
### Summary

The research entrypoint accepts a caller-supplied `source_urls` list over the unauthenticated `/ws` WebSocket. Any entry ending in `.pdf` is routed to `PyMuPDFScraper`, whose non-URL branch passes the value to `PyMuPDFLoader` as a local filesystem path, reads the PDF, and returns its parsed text in the streamed context and final report. Because a bare local path has no URL scheme/netloc, it takes the local-file branch, giving an unauthenticated network attacker arbitrary local PDF file read with content readback. Confirmed against the real scraper: a server-side secret PDF was read and its text returned. Note this shares the `source_urls`/`/ws` input with the published SSRF advisory but is a distinct sink (local file read, CWE-22/73 vs CWE-918).

### Details
based on commit b364917
`gpt_researcher/scraper/pymupdf/pymupdf.py`:

```python
def is_url(self):                       # ~21-32
    return all([scheme, netloc])        # bare path -> False
...
else:
    loader = PyMuPDFLoader(self.link)   # ~65, local-file branch
    return loader.load()
```

Router: `gpt_researcher/scraper/scraper.py:201` -- `if link.endswith(".pdf"): scraper_key = "pdf"`, so any `.pdf`-suffixed string is routed to `PyMuPDFScraper` regardless of the configured default scraper (`bs`). A bare local path like `/var/app/secret.pdf` has empty scheme/netloc, so `is_url()` is False and the local-file `else` branch runs.

Data flow: unauthenticated `/ws` client (`backend/server/app.py:362`) sends `start {...}` -> `extract_command_data` (`server_utils.py:398`) reads attacker `source_urls` -> `GPTResearcher.source_urls` (`researcher.py:137` `_get_context_by_urls` -> `_get_new_urls` `researcher.py:728`, no scheme/path validation) -> `scraper_manager.browse_urls` -> `actions/web_scraping.py:34` `Scraper` -> `get_scraper(link)` -> `PyMuPDFScraper`. The parsed local PDF content is returned to the attacker in the streamed research context, the final report, and the downloadable report artifact.

The boundary crossed is an unauthenticated network attacker causing the server to read arbitrary local PDF files and disclose their contents (other users' uploads in `DOC_PATH`/`./my-docs`, generated reports in `outputs/`, any host PDF).

the target path must literally end in `.pdf` and be a genuine PDF (the loader validates filetype from the extension), so this is arbitrary PDF read, not unconstrained file read. The input parameter and endpoint match the published SSRF advisory (CVE-2026-5633 / GHSA-8p84-j4x4-c993), but that PoC is CWE-918 http/https SSRF; this is the never-validated local-file `else` branch (CWE-22/73), not covered by it. A strict triager may still treat it as the same `source_urls` root cause.

### PoC
<available upon request>

### Impact

An unauthenticated client of an exposed gpt-researcher API/`/ws` reads arbitrary local PDF files and receives their contents in the research output: other users' uploaded documents, generated reports, and any PDF on the host. Real-world exposure depends on the operator exposing the API (common in hosted/team mode); default deployments bind locally.

### Remediation

In `PyMuPDFScraper.scrape()` and the `get_scraper` dispatch, reject non-HTTP(S) inputs before the local-file branch: require `urlparse(link).scheme in ("http","https")` and drop the `else: PyMuPDFLoader(self.link)` local-path path for network-sourced URLs. More robustly, add a single validation gate on every entry in `source_urls`/`document_urls` in `_get_new_urls` (block non-`http(s)` schemes, bare paths, and SSRF-range hosts), which also closes the SSRF root cause.

VenkateswarluNagineni:
## Summary

User-supplied `source_urls` and `document_urls` (accepted by the API and WebSocket endpoints) were fetched without any validation. This allowed two unauthenticated attacks:

- **SSRF (#1794):** an attacker could point the researcher at internal services, private-network hosts, or the cloud metadata endpoint (`http://169.254.169.254/...`) and have the response embedded in the generated report.
- **Arbitrary local file read (#1805):** a non-URL path (or `file://` URL) reaching `PyMuPDFLoader`/`PyMuPDFScraper`'s local-file branch caused the server to read files off its own disk.

Both stem from the same root cause: untrusted URLs flowing into the scraping / document-loading pipelines unchecked.

## Changes

- Add `gpt_researcher/utils/url_security.py` with `validate_url()` / `is_safe_url()`:
  - allow only `http` / `https` schemes (blocks `file://`, `ftp://`, `gopher://`, …);
  - require a host (blocks bare local filesystem paths);
  - resolve the host and reject **non-public** addresses — private, loopback, link-local, reserved, multicast and unspecified — which also covers the `169.254.169.254` metadata endpoint.
- Enforce it at the two chokepoints every untrusted URL passes through:
  - `Scraper.extract_data_from_url` (covers `source_urls` → all scraper backends, incl. the PyMuPDF local-file branch);
  - `OnlineDocumentLoader._download_and_process` (covers `document_urls`).
- Unsafe URLs are skipped with a warning rather than crashing the run.

## Opt-out for trusted / self-hosted deployments

Operators who intentionally scrape internal resources can set `ALLOW_PRIVATE_URLS=true` (or pass `allow_private=True`) to disable the private-address check. Scheme/host validation still applies.

## Tests

Adds `tests/test_url_security.py` (21 cases): disallowed schemes and local paths, internal/loopback/link-local/metadata IPv4 + IPv6 targets, allowed public IPs, a DNS-rebinding case (public-looking host resolving to a private IP), and the env-var / argument opt-out.

## Note on scope

Validation resolves the host and then lets the HTTP client re-resolve it, leaving a small TOCTOU / DNS-rebinding window. Fully closing it requires pinning the validated IP into the connection. This PR is intended as defence-in-depth that removes the trivial, unauthenticated SSRF and local-file-read primitives; pinning can be a follow-up.

Fixes #1794
Fixes #1805

### DIFF
--- a/gpt_researcher/document/online_document.py
+++ b/gpt_researcher/document/online_document.py
@@ -1,6 +1,8 @@
 import os
 import aiohttp
 import tempfile
+
+from gpt_researcher.utils.url_security import UnsafeURLError, validate_url
 from langchain_community.document_loaders import (
     PyMuPDFLoader,
     TextLoader,
@@ -35,6 +37,13 @@ class OnlineDocumentLoader:
 
     async def _download_and_process(self, url: str) -> list:
         try:
+            # Reject SSRF / local-file targets before issuing the request.
+            try:
+                validate_url(url)
+            except UnsafeURLError as e:
+                print(f"Skipping unsafe document URL {url}: {e}")
+                return []
+
             headers = {
                 "User-Agent": "Mozilla/5.0"
             }
--- a/gpt_researcher/scraper/scraper.py
+++ b/gpt_researcher/scraper/scraper.py
@@ -14,6 +14,7 @@ import requests
 from colorama import Fore, init
 
 from gpt_researcher.utils.workers import WorkerPool
+from gpt_researcher.utils.url_security import UnsafeURLError, validate_url
 
 from . import (
     ArxivScraper,
@@ -111,6 +112,19 @@ class Scraper:
         """
         async with self.worker_pool.throttle():
             try:
+                # Reject SSRF / local-file targets (internal hosts, cloud metadata
+                # endpoints, file:// paths, etc.) before any request is made.
+                try:
+                    validate_url(link)
+                except UnsafeURLError as e:
+                    self.logger.warning(f"Skipping unsafe URL {link}: {e}")
+                    return {
+                        "url": link,
+                        "raw_content": None,
+                        "image_urls": [],
+                        "title": "",
+                    }
+
                 Scraper = self.get_scraper(link)
                 scraper = Scraper(link, session)
 
--- a/gpt_researcher/utils/url_security.py
+++ b/gpt_researcher/utils/url_security.py
@@ -1,0 +1,125 @@
+"""URL security utilities for GPT Researcher.
+
+These helpers protect the scraping and document-loading pipelines against
+Server-Side Request Forgery (SSRF) and arbitrary local-file reads when the URLs
+originate from untrusted input (e.g. the ``source_urls`` / ``document_urls``
+parameters accepted by the API and WebSocket endpoints).
+
+By default only ``http`` / ``https`` URLs whose host resolves to a public IP
+address are allowed. This blocks:
+
+* Non-HTTP schemes such as ``file://``, ``ftp://`` and ``gopher://``.
+* Local filesystem paths (no scheme/host), which some document loaders would
+  otherwise read directly from disk (arbitrary local file read).
+* Requests to private, loopback, link-local, reserved or otherwise internal
+  addresses (e.g. ``127.0.0.1``, ``10.0.0.0/8`` and the ``169.254.169.254``
+  cloud metadata endpoint).
+
+Operators who intentionally scrape internal or self-hosted resources can opt out
+of the private-address check by setting the environment variable
+``ALLOW_PRIVATE_URLS=true`` (or by passing ``allow_private=True``).
+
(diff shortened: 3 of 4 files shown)


assafelovic:

**Note:** this changes behaviour for anyone scraping internal hosts — the guard resolves each URL and rejects private, loopback, link-local and metadata addresses. `ALLOW_PRIVATE_URLS` is now documented in `.env.example` as the opt-out.

FlorianFranzen:
@VenkateswarluNagineni Please be aware that despite what the LLM generated comment wants you to believe, your change has not been attributed with the appropriate `Co-Authored-By:` byline when it was merged upstream.

VenkateswarluNagineni:
Thanks for flagging this, @FlorianFranzen — you're right, the `Co-Authored-By:` trailer on `39215c5b` only lists the AI tooling, not the human contributors whose PRs were bundled in (including #1818). The PR body does reference `#1818` by number so the change is traceable either way, so this is mostly a nice-to-have. @assafelovic, if it's easy to add the human `Co-Authored-By:` trailers in a follow-up commit that'd be appreciated, but totally understand if that's not practical for a squash this size — no worries either way!

PR #1823, opened 2026-06-27
fix(query): always return list[str] from generate_sub_queries

Bartok9:
## Summary

- `generate_sub_queries` returned the raw output of `json_repair.loads`, which is not guaranteed to be a list.
- Callers treat it as `list[str]` and crash when the model returns a dict, a bare string, or unparseable output.

## Root Cause

**Symptom** — A research run aborts with `AttributeError: 'str' object has no attribute 'append'` (or iterates a dict's keys instead of queries) when the strategic/smart LLM does not emit a clean JSON array of sub-queries.

**Root cause** — `generate_sub_queries()` in `gpt_researcher/actions/query_processing.py` ended with `return json_repair.loads(response)`. `json_repair.loads` returns whatever shape it can recover:

```
'{"queries": ["a","b"]}' -> {'queries': ['a', 'b']}   (dict)
'not json at all'        -> ''                          (str)
'["x","y"]'              -> ['x', 'y']                   (list)
'{"q":"single"}'         -> {'q': 'single'}             (dict)
''                       -> ''                           (str)
```

But `ResearchConductor.plan_research` consumers assume a list:

```python
sub_queries = await self.plan_research(query)          # researcher.py:242
if self.researcher.report_type != "subtopic_report":
    sub_queries.append(query)                          # crashes if str/dict
...
for sub_query in sub_queries:                           # iterates dict keys, not queries
```

**Evidence** — The `json_repair` shapes above are reproduced directly against the pinned dependency; the `.append` / iteration call sites are at `gpt_researcher/skills/researcher.py:242-261` and `:330`.

**Fix + why this level** — Add `_normalize_sub_queries(parsed, fallback_query)` and return it from `generate_sub_queries`. It coerces dict (`queries`/`sub_queries`/`subQueries`/`items` or a single `query`), bare string, and `None`/`""` into a flat `list[str]`, falling back to the original query so research never proceeds with an empty plan. Fixing at the producer guarantees the documented `List[str]` return contract for every caller, instead of scattering shape-guards across call sites.

**Scope / risk** — One helper + one return line, plus a new test module. The common, already-working case (a JSON array) is unchanged. Non-list responses that previously crashed now degrade to a usable query list.

## Verification

```
python -m pytest tests/test_sub_query_normalization.py -q
8 passed

python -m pytest tests/test_deep_research_parsing.py tests/test_costs.py -q
29 passed
```

New `tests/test_sub_query_normalization.py` covers list, dict-with-queries-key, single-query dict, bare string, empty string, `None`, blank-stripping, and unrecognized dict shape (all of which previously had no guarantee of a list result).

## What was NOT changed

- The LLM call/fallback chain (strategic → token-limit retry → smart) is untouched.
- JSON array responses produce identical output to before.

### DIFF
--- a/gpt_researcher/actions/query_processing.py
+++ b/gpt_researcher/actions/query_processing.py
@@ -1,6 +1,37 @@
 import json_repair
 
 from gpt_researcher.llm_provider.generic.base import ReasoningEfforts
+
+
+def _normalize_sub_queries(parsed: Any, fallback_query: str) -> List[str]:
+    """Coerce a parsed LLM response into a flat list of query strings.
+
+    ``json_repair.loads`` may return a list, a dict (e.g. ``{"queries": [...]}``
+    or a single ``{"query": "..."}``), a bare string, or ``None`` when the model
+    does not return clean JSON. Callers expect a ``list[str]`` and otherwise crash
+    on ``.append`` / iteration, so normalize defensively here.
+    """
+    if isinstance(parsed, dict):
+        for key in ("queries", "sub_queries", "subQueries", "items"):
+            value = parsed.get(key)
+            if isinstance(value, list):
+                parsed = value
+                break
+        else:
+            # Single-query dict like {"query": "..."} or unrecognized shape.
+            single = parsed.get("query")
+            parsed = [single] if isinstance(single, str) else []
+
+    if isinstance(parsed, str):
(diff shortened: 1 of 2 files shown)


assafelovic:
Superseded by #1910 ✅

PR #1832, opened 2026-06-28
fix(scraper): detect PDF links with query strings/uppercase in BrowserScraper

Bartok9:
## Summary

- `BrowserScraper` (Selenium) routed pages to its PDF handler with `self.url.endswith(".pdf")`, which missed PDFs behind a query string/fragment and uppercase `.PDF`.
- Such links silently fell through to HTML scraping and returned non-PDF garbage.

## Motivation

Signed CDN/S3 PDF links almost always carry a signature query string (`https://host/report.pdf?sig=...&exp=...`), and uppercase `.PDF` is a valid suffix. The bare suffix check failed on both, so the browser scraper never invoked `scrape_pdf_with_pymupdf` for them.

This mirrors the same class of bug already fixed for `Scraper.get_scraper` in #1828 — but the `BrowserScraper` path is independent and was untouched.

## Fix

Extracted a small `_is_pdf_url(url)` helper that inspects only `urlparse(url).path`, lowercased, before checking the `.pdf` suffix. The scrape path now calls it instead of `endswith`.

## Real behavior proof

**Behavior addressed:** `endswith(".pdf")` returns `False` for `doc.pdf?sig=x` and `R.PDF`.

**Exact commands run after the patch:**
```
python -m pytest tests/test_browser_pdf_detection.py -q
```

**Captured output AFTER fix:**
```
7 passed, 2 warnings in 0.34s
```

**Pre-fix demonstration (old logic):**
```
query: False   # https://cdn/doc.pdf?sig=x  -> should be a PDF
upper: False   # https://x/R.PDF            -> should be a PDF
```

**Regression test:** `tests/test_browser_pdf_detection.py` — asserts query-string, fragment, and uppercase PDFs are detected and that a `.pdf` appearing only in a query param is NOT. Fails without the fix.

## Scope / risk

Pure detection helper; only the browser scrape branch changed. Non-PDF and plain `.pdf` behavior unchanged.

### DIFF
--- a/gpt_researcher/scraper/browser/browser.py
+++ b/gpt_researcher/scraper/browser/browser.py
@@ -15,7 +15,20 @@ from typing import Iterable, cast
 from .processing.scrape_skills import (scrape_pdf_with_pymupdf,
                                        scrape_pdf_with_arxiv)
 
-from urllib.parse import urljoin
+from urllib.parse import urljoin, urlparse
+
+
+def _is_pdf_url(url: str) -> bool:
+    """Return True when ``url`` points at a PDF.
+
+    Inspects only the path component so query strings / fragments don't hide
+    the extension (signed CDN/S3 links like ``https://host/doc.pdf?sig=...``
+    are extremely common), and matches case-insensitively because ``.PDF`` is
+    a perfectly valid suffix.
+    """
+    if not url:
+        return False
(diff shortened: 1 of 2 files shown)


assafelovic:
Superseded by #1908 ✅

PR #1833, opened 2026-06-28
fix(retriever): preserve camelCase sort criterion for Semantic Scholar

Bartok9:
## Summary

- `SemanticScholarSearch.__init__` validated the `sort` argument against camelCase API values, then stored `sort.lower()` — corrupting `citationCount` → `citationcount` and `publicationDate` → `publicationdate`.
- The lowercased value was sent as the `sort` query parameter; the Semantic Scholar API expects the exact camelCase form, so non-default sorting was silently broken.

## Motivation

```python
VALID_SORT_CRITERIA = ["relevance", "citationCount", "publicationDate"]
...
assert sort in self.VALID_SORT_CRITERIA, "Invalid sort criterion"
self.sort = sort.lower()   # <- corrupts the just-validated value
```

The assert guarantees `sort` is already a valid API criterion; lowercasing it immediately afterward makes the stored value no longer a member of `VALID_SORT_CRITERIA`. `relevance` happened to survive (already lowercase), masking the bug for the default.

## Fix

Store the validated value verbatim (`self.sort = sort`).

## Real behavior proof

**Behavior addressed:** stored sort no longer matches the API's accepted set.

**Exact commands run:**
```
python -m pytest tests/test_semantic_scholar_sort.py -q
```

**Captured output AFTER fix:**
```
4 passed, 2 warnings in 0.30s
```

**Captured output BEFORE fix (test stashed back onto unpatched source):**
```
3 failed, 1 passed
FAILED tests/test_semantic_scholar_sort.py::test_camelcase_sort_preserved
FAILED tests/test_semantic_scholar_sort.py::test_publication_date_sort_preserved
FAILED tests/test_semantic_scholar_sort.py::test_stored_sort_is_a_valid_api_value
```

**Regression test:** `tests/test_semantic_scholar_sort.py` — asserts the stored `sort` stays camelCase and remains a member of `VALID_SORT_CRITERIA`. Fails without the fix.

## Scope / risk

One assignment changed. `relevance` (the default) is unaffected; only the two camelCase criteria are corrected.

### DIFF
--- a/gpt_researcher/retrievers/semantic_scholar/semantic_scholar.py
+++ b/gpt_researcher/retrievers/semantic_scholar/semantic_scholar.py
@@ -20,7 +20,11 @@ class SemanticScholarSearch:
         """
         self.query = query
         assert sort in self.VALID_SORT_CRITERIA, "Invalid sort criterion"
-        self.sort = sort.lower()
+        # Preserve the exact (camelCase) criterion. The Semantic Scholar API
+        # expects ``citationCount`` / ``publicationDate`` verbatim; lowercasing
+        # them produced ``citationcount`` / ``publicationdate``, which the API
+        # rejects or silently ignores.
+        self.sort = sort
 
     def search(self, max_results: int = 20) -> List[Dict[str, str]]:
         """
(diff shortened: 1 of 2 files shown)


assafelovic:
Superseded by #1907 ✅

PR #1834, opened 2026-06-28
fix(agent): capture whole JSON object in extract_json_with_regex

Bartok9:
## Summary

- `extract_json_with_regex` (the fallback JSON recovery used when an LLM agent-selection response fails strict parsing) used a non-greedy `{.*?}` pattern.
- Non-greedy stops at the **first** `}`, so any object with more than one key, or with a `}` inside a string value, was truncated into an invalid fragment that `json.loads` then rejected — silently dropping the agent the LLM chose and falling back to the Default Agent.

## Motivation

```python
re.search(r"{.*?}", response, re.DOTALL)
```

Real responses look like:
```
Here is the agent: {"server": "Finance", "agent_role_prompt": "analyze {markets} data"} done
```
The non-greedy match returns:
```
{"server": "Finance", "agent_role_prompt": "analyze {markets}
```
…which is invalid JSON. Even a plain two-key object (`{"server": "A", "agent_role_prompt": "B"}`) is truncated at the first `}` whenever one exists earlier — and agent role prompts very often contain braces.

## Fix

Use a greedy `{.*}` so the match spans from the first `{` to the **last** `}`, capturing the whole object.

## Real behavior proof

**Commands:**
```
python -m pytest tests/test_extract_json_with_regex.py -q
```

**AFTER fix:**
```
4 passed, 2 warnings in 0.31s
```

**BEFORE fix (tests stashed onto unpatched source):**
```
1 failed, 3 passed
FAILED tests/test_extract_json_with_regex.py::test_brace_inside_string_value_preserved
```

**Captured helper output AFTER fix:**
```
{'server': 'Finance', 'agent_role_prompt': 'analyze {markets} data'}
{'server': 'A', 'agent_role_prompt': 'B'}
```

## Regression test

`tests/test_extract_json_with_regex.py` — asserts multi-key objects and brace-in-string-value objects round-trip through `json.loads`. The brace-in-value test fails without the fix.

## Scope / risk

One regex literal. `None`/no-JSON inputs still return `None` (covered). A greedy match across multiple JSON objects in one string would over-capture, but this helper targets the single agent-selection object the LLM emits.

### DIFF
--- a/gpt_researcher/actions/agent_creator.py
+++ b/gpt_researcher/actions/agent_creator.py
@@ -120,7 +120,12 @@ def extract_json_with_regex(response: str | None) -> str | None:
     """
     if not response:
         return None
-    json_match = re.search(r"{.*?}", response, re.DOTALL)
+    # Greedy ``{.*}`` so the match spans from the first ``{`` to the LAST ``}``
+    # in the response, capturing the whole object. A non-greedy ``{.*?}``
+    # stopped at the first ``}``, truncating any object with more than one
+    # key or with a ``}`` inside a string value (e.g. an agent_role_prompt
+    # mentioning "{markets}") into invalid JSON.
+    json_match = re.search(r"{.*}", response, re.DOTALL)
     if json_match:
         return json_match.group(0)
     return None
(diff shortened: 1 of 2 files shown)


assafelovic:
Superseded by #1910 ✅

PR #1835, opened 2026-06-29
feat(agent): add all_retrievers option to quick_search

Bartok9:
## Problem

`GPTResearcher.quick_search()` is documented as a lightweight web search, but it only ever queries **the first configured retriever** (`self.retrievers[0]`). When a user configures multiple retrievers — e.g. `RETRIEVER="tavily,duckduckgo"` — `quick_search` silently ignores all but the first. Full research (`conduct_research`) fans out across retrievers, so this is an inconsistent and surprising gap for anyone using `quick_search` as a fast multi-source lookup or chat tool.

## Solution

Add an opt-in `all_retrievers` flag to `quick_search`:

- When `all_retrievers=True` (and more than one retriever is configured), it queries **every** configured retriever concurrently via `asyncio.gather` and merges the results.
- Results are **de-duplicated by URL**, handling both `url` and `href` keys (different retrievers use different shapes — Tavily returns `href`, others `url`).
- A retriever that raises is **skipped** (`return_exceptions=True`), so one failing provider doesn't abort the whole search.
- Default behaviour is **unchanged** (`all_retrievers=False` → primary retriever only), so this is fully backward compatible.

Also threads the `researcher=self` argument through the primary-path `get_search_results` call so MCP retrievers work in single-retriever quick searches too.

## Files changed

- `gpt_researcher/agent.py` — new `all_retrievers` param + `_search_all_retrievers` helper.
- `tests/test_quick_search.py` — 3 new tests (merge+dedup, failing-retriever skip, single-retriever fallback).

## Tests

```
$ python -m pytest tests/test_quick_search.py -v
...
5 passed, 3 warnings in 0.66s
```

All existing quick_search tests still pass; new tests cover the merge/dedup, fault-tolerance, and backward-compat paths.

## Breaking changes

None. The new parameter defaults to `False`.

### DIFF
--- a/gpt_researcher/agent.py
+++ b/gpt_researcher/agent.py
@@ -4,6 +4,7 @@ This module provides the main GPTResearcher class that orchestrates
 autonomous research and report generation using LLMs and web search.
 """
 
+import asyncio
 import json
 import os
 from typing import Any, Optional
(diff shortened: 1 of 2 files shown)


assafelovic:
Superseded by #1910 ✅

PR #1837, opened 2026-06-29
fix(retriever): return empty list instead of None from SerperSearch.search

Bartok9:
## Summary

- `SerperSearch.search` now returns `[]` (not `None`) on its three error paths.
- Prevents `TypeError: object of type 'NoneType' has no len()` when a Serper request fails.

## Motivation

`SerperSearch.search` documents a `list` return type, and every sibling
retriever (`serpapi`, `brave`, `bing`, `searx`) returns `[]` on failure.
But three paths returned a bare `None`:

```python
if resp is None:
    return            # -> None
try:
    search_results = json.loads(resp.text)
except Exception:
    return            # -> None
if search_results is None:
    return            # -> None
```

Callers treat the result as a list. `get_search_results` is annotated
`-> List[Dict[str, Any]]`, and `skills/researcher.py` does this immediately:

```python
search_results = await get_search_results(query, self.researcher.retrievers[0], ...)
self.logger.info(f"Initial search results obtained: {len(search_results)} results")
```

So a transient Serper failure (network blip, rate-limit HTML body, `null`
JSON) crashed the whole research run with
`TypeError: object of type 'NoneType' has no len()` instead of degrading to
an empty result set the way the other retrievers do.

## Verification

- `python -m pytest tests/test_serper_returns_list.py` — 3 passed.

## Real behavior proof

Regression test drives all three error paths with a mocked `requests.request`.

**Without the fix:**
```
FAILED tests/test_serper_returns_list.py::test_search_returns_list_when_response_none
FAILED tests/test_serper_returns_list.py::test_search_returns_list_when_body_unparseable
FAILED tests/test_serper_returns_list.py::test_search_returns_list_when_json_null
3 failed
```
Each assertion `len(out) == 0` raises `TypeError: object of type 'NoneType' has no len()` — the exact production failure.

**With the fix:** `3 passed`.

## Scope / risk

Three one-word changes (`return` -> `return []`) in the error branches only.
The success path is untouched, so well-formed responses behave identically.

### DIFF
--- a/gpt_researcher/retrievers/serper/serper.py
+++ b/gpt_researcher/retrievers/serper/serper.py
@@ -104,15 +104,16 @@ class SerperSearch():
 
         resp = requests.request("POST", url, timeout=10, headers=headers, data=data)
 
-        # Preprocess the results
+        # Preprocess the results. Always return a list so callers (which do
+        # `len(...)` / iterate over the result) never receive None.
         if resp is None:
-            return
+            return []
         try:
             search_results = json.loads(resp.text)
         except Exception:
-            return
+            return []
         if search_results is None:
-            return
+            return []
 
         results = search_results.get("organic", [])
         search_results = []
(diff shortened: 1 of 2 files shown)


assafelovic:
Superseded by #1907 ✅

PR #1838, opened 2026-06-29
fix(retriever): return empty list instead of None from PubMedCentralSearch.search

Bartok9:
## Summary

- `PubMedCentralSearch.search` now returns `[]` (not `None`) when no articles are found.
- Prevents `TypeError: object of type 'NoneType' has no len()` on a zero-hit PubMed query.

## Motivation

`search` returned `None` whenever `_search_articles` produced nothing — a
genuine search miss, or a transient esearch failure:

```python
article_ids = self._search_articles(max_results)
if not article_ids:
    return None
```

But the result is consumed as a list. `actions.query_processing.get_search_results`
is annotated `-> List[Dict[str, Any]]`, and `skills/researcher.py` runs:

```python
search_results = await get_search_results(query, self.researcher.retrievers[0], ...)
self.logger.info(f"Initial search results obtained: {len(search_results)} results")
```

So a query that returns no PubMed articles crashed the research run with
`TypeError: object of type 'NoneType' has no len()`, instead of degrading to
an empty result set like every other retriever (serpapi/brave/bing/searx all
return `[]`).

## Verification

- `python -m pytest tests/test_pubmed_central_returns_list.py` — 2 passed.

## Real behavior proof

Regression test patches `_search_articles` to return `None` and `[]`.

**Without the fix:**
```
FAILED tests/test_pubmed_central_returns_list.py::test_search_returns_list_when_no_article_ids
FAILED tests/test_pubmed_central_returns_list.py::test_search_returns_list_when_empty_article_ids
2 failed
```
The `len(out) == 0` assertion raises `TypeError: object of type 'NoneType' has no len()` — the exact production failure.

**With the fix:** `2 passed`.

## Scope / risk

One return path changed (`return None` -> `return []`) plus a return-type
annotation tightened to `List[Dict[str, Any]]`. The success path (articles
found) is untouched.

### DIFF
--- a/gpt_researcher/retrievers/pubmed_central/pubmed_central.py
+++ b/gpt_researcher/retrievers/pubmed_central/pubmed_central.py
@@ -125,7 +125,7 @@ class PubMedCentralSearch:
         except requests.RequestException as e:
             return None
 
-    def search(self, max_results: int = 5) -> Optional[List[Dict[str, Any]]]:
+    def search(self, max_results: int = 5) -> List[Dict[str, Any]]:
         """
         Performs the search and retrieves full text content.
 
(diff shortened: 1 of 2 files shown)


assafelovic:
Superseded by #1907 ✅

PR #1839, opened 2026-06-29
fix(scraper): guard ArxivScraper against empty retriever results

Bartok9:
## Summary

- `ArxivScraper.scrape` now returns `("", [], "")` when the retriever yields no documents instead of raising `IndexError`.
- Metadata is read with `.get()` so a paper missing `Published`/`Authors`/`Title` no longer raises `KeyError`.

## Motivation

`scrape` indexed the first document unconditionally:

```python
docs = retriever.invoke(query)
context = f"Published: {docs[0].metadata['Published']}; Author: {docs[0].metadata['Authors']}; Content: {docs[0].page_content}"
...
return context, image, docs[0].metadata["Title"]
```

`ArxivRetriever` returns an **empty list** when the query (derived from the
URL's last path segment) matches no paper — a malformed or non-arXiv link, or
a transient arXiv API failure. `docs[0]` then raised `IndexError` and aborted
the entire scrape run, even though every other scraper (`BeautifulSoupScraper`,
`WebBaseLoaderScraper`, `FireCrawl`) degrades gracefully to `("", [], "")`.

The hard-keyed `metadata['Published'|'Authors'|'Title']` access was a second
latent crash for any record missing one of those fields.

## Verification

- `python -m pytest tests/test_arxiv_scraper_empty.py` — 2 passed.

## Real behavior proof

Regression test mocks `ArxivRetriever` to return `[]` and to return a doc.

**Without the fix:**
```
FAILED tests/test_arxiv_scraper_empty.py::test_scrape_returns_empty_when_no_docs
E       IndexError: list index out of range
1 failed, 1 passed
```
— the exact production crash.

**With the fix:** `2 passed` (empty -> `("", [], "")`; present -> content + title).

## Scope / risk

Adds an empty-list guard and switches metadata reads to `.get()`. The happy
path (a paper is found with full metadata) is behaviourally unchanged.

### DIFF
--- a/gpt_researcher/scraper/arxiv/arxiv.py
+++ b/gpt_researcher/scraper/arxiv/arxiv.py
@@ -37,16 +37,24 @@ class ArxivScraper:
 
         Returns:
             (context, images, title) matching other scrapers.
+
+        When the query matches no paper (malformed/non-arXiv id, or empty
+        client results), degrade to an empty result instead of raising so a
+        single bad URL cannot abort the whole scrape pipeline.
         """
+        paper_id = _paper_id_from_link(self.link)
+        if not paper_id:
+            return "", [], ""
+
         import arxiv
 
-        paper_id = _paper_id_from_link(self.link)
         client = arxiv.Client()
         search = arxiv.Search(id_list=[paper_id], max_results=1)
         try:
             paper = next(client.results(search))
-        except StopIteration as exc:
-            raise ValueError(f"No arXiv paper found for id {paper_id!r}") from exc
+        except StopIteration:
+            # No matching paper — mirror other scrapers: empty degrade.
+            return "", [], ""
 
         authors = ", ".join(a.name for a in (paper.authors or []))
         published = paper.published.date().isoformat() if paper.published else ""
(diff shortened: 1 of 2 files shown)


assafelovic:
Thanks @Bartok9! The arXiv scraper was rewritten to use the Client API in #1906 (your #1896), which touches the same code this fix guards. This now conflicts — could you rebase onto the latest `main` so the empty-results guard sits on top of the new arXiv fetch? 🙏

Bartok9:
🤖 Ack — #1906 rewrites this scraper to the `arxiv.Client` API and changes the no-paper path to raise `ValueError` (vs my degrade-to-empty). Rather than rebase a guard that the rewrite reshapes, I'd rather not create churn on top of an unmerged PR. Once #1906 lands, if you still want the no-paper case to degrade gracefully instead of raising, ping me and I'll add a focused follow-up on the new code. Otherwise happy to close as superseded — your call. 🙏

Bartok9:
@assafelovic Rebased onto latest `main` (`c8604e1f`). Empty-results guard is back on the current `ArxivRetriever` path (`if not docs: return \"\", [], \"\"` + safe metadata `.get`). Note: Client-API rewrite in open #1906 still reshapes this path to raise on missing paper; once that lands we can re-follow if you want degrade-to-empty on the new API. Ready for re-review. 🙏

assafelovic:

Bartok9:
Thanks @assafelovic — glad the empty-results guard landed via #2074 (`9140bde5`) with co-author credit. Appreciate you stacking and verifying across 3.12–3.14.

PR #1842, opened 2026-06-30
fix(scraper): remove PyMuPDF temp file even when PDF loading fails

Bartok9:
## Summary

- `PyMuPDFScraper.scrape` downloads a remote PDF into a `NamedTemporaryFile(delete=False, suffix=".pdf")` and then loads it with `PyMuPDFLoader`. It called `os.remove(temp_filename)` **only on the success path**.
- User-facing impact: every time `PyMuPDFLoader.load()` raises (malformed / partial / truncated PDF, common with flaky downloads), the broad `except` swallows the error and the downloaded temp `.pdf` is left behind in the temp dir. Over a long run that's an unbounded disk leak.

## Motivation

The fix wraps the load in `try/finally` so the temp file is removed whether or not loading succeeds (the `os.remove` itself is guarded with `except OSError` so cleanup can't mask the real error). The empty-result contract `("", [], "")` on failure is unchanged.

## Verification

- `python -m pytest tests/test_pymupdf_tempfile_cleanup.py -q` — 1 passed.

## Real behavior proof

**Regression test** `tests/test_pymupdf_tempfile_cleanup.py` mocks `requests.get` to return PDF bytes and `PyMuPDFLoader.load` to raise `RuntimeError("corrupt PDF")`, then asserts no new `*.pdf` appears in the temp dir.

Without the fix:
```
FAILED tests/test_pymupdf_tempfile_cleanup.py::test_tempfile_removed_when_loader_raises
1 failed
```
(a leaked temp file remains)

With the fix:
```
tests/test_pymupdf_tempfile_cleanup.py .                                 [100%]
1 passed
```

## Scope / risk

- Touches only `gpt_researcher/scraper/pymupdf/pymupdf.py` (the URL download branch) + a new test.
- Local-file branch (no temp file) and the empty-result error contract are unchanged.

### DIFF
--- a/gpt_researcher/scraper/pymupdf/pymupdf.py
+++ b/gpt_researcher/scraper/pymupdf/pymupdf.py
@@ -57,10 +57,16 @@ class PyMuPDFScraper:
                     for chunk in response.iter_content(chunk_size=8192):
                         temp_file.write(chunk)  # Write the downloaded content to the temporary file
 
-                loader = PyMuPDFLoader(temp_filename)
-                doc = loader.load()
-
-                os.remove(temp_filename)
+                # Always clean up the downloaded temp file, even if loading fails
+                # (PyMuPDFLoader.load() can raise on a malformed/partial PDF).
+                try:
+                    loader = PyMuPDFLoader(temp_filename)
+                    doc = loader.load()
+                finally:
+                    try:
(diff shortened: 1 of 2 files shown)


assafelovic:
Superseded by #1908 ✅

PR #1843, opened 2026-06-30
fix(deep-research): don't drop all context when newest item exceeds word limit

Bartok9:
## Summary

- `trim_context_to_word_limit` keeps the most recent context items within a word budget. It iterated newest-first and `break`ed on the first item that didn't fit.
- User-facing impact: when the **most recent** context item alone exceeds `max_words` (a long scraped page / big learning), the `break` fires on the *first* iteration and discards the **entire** context list — including smaller, older items that would have fit comfortably. Deep research then proceeds with zero context for that step.

## Motivation

Replace the unconditional `break` with: skip an oversized item and keep scanning; `break` only once the budget is actually exhausted. Smaller items still get included even if a newer item was too big.

## Verification

- `python -m pytest tests/test_trim_context_word_limit.py -q` — 3 passed.

## Real behavior proof

**Repro (old code):**
```
items = ['small one', 'small two', 'x '*5000]   # newest is huge
trim_context_to_word_limit(items, max_words=100)  ->  []   # everything dropped
```

**Regression test** `tests/test_trim_context_word_limit.py` asserts the two small items survive when the newest item is oversized. It FAILS without the fix:
```
FAILED tests/test_trim_context_word_limit.py::test_oversized_recent_item_does_not_drop_smaller_older_items
1 failed, 2 passed
```
and passes with it:
```
3 passed
```
The other two tests confirm in-budget items are all kept in order and the budget still stops once exhausted.

## Scope / risk

- Touches only `trim_context_to_word_limit` in `gpt_researcher/skills/deep_research.py` + a new test.
- In-budget behavior and ordering are unchanged (covered by tests).

### DIFF
--- a/gpt_researcher/skills/deep_research.py
+++ b/gpt_researcher/skills/deep_research.py
@@ -32,7 +32,12 @@ def trim_context_to_word_limit(context_list: List[str], max_words: int = MAX_CON
             trimmed_context.insert(0, item)  # Insert at start to maintain original order
             total_words += words
         else:
-            break
+            # Skip this item but keep scanning: a single oversized entry (often
+            # the most recent one) must not drop every smaller item that would
+            # still fit. Stop only once the budget is exhausted.
+            if total_words >= max_words:
+                break
+            continue
 
     return trimmed_context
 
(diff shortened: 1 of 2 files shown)


assafelovic:

Bartok9:
🤖 Confirmed covered — `trim_context_to_word_limit` on current `main` already keeps the oversized newest item (truncated to the word limit) instead of dropping all context (`deep_research.py`, the `elif not trimmed_context` branch). Closing as covered. Thanks! 🙏

PR #1844, opened 2026-06-30
fix(retriever): handle errors and missing keys in BoChaSearch

Bartok9:
## Summary

- `BoChaSearch.search` indexed `json_response["data"]["webPages"]["value"]` directly and had **no** error handling around the HTTP call or the response shape.
- User-facing impact: a non-200 response (rate limit / bad key / server error), a body that doesn't decode as JSON, or a payload missing those keys raises an unhandled `KeyError` / `JSONDecodeError` that propagates out of `search()` and crashes the whole research run. Every other retriever in the repo (serper, searchapi, semantic_scholar, etc.) degrades to `[]` on failure.

## Motivation

Make BoCha consistent with its sibling retrievers: add `timeout` + `raise_for_status()`, catch `requests.RequestException` / `ValueError` (JSON decode) and return `[]`, walk `data.webPages.value` defensively, and use `result.get(...)` so an item missing a field doesn't crash the loop.

> Note: a prior community PR (#1722) raised the same issue but was closed without a maintainer review and shipped no test. This version adds full regression coverage.

## Verification

- `python -m pytest tests/test_bocha_error_handling.py -q` — 3 passed.

## Real behavior proof

**Regression tests** `tests/test_bocha_error_handling.py` cover (1) a payload missing the expected keys, (2) a `requests.RequestException`, and (3) a valid payload (including a result missing `snippet`). All three FAIL without the fix:
```
FAILED tests/test_bocha_error_handling.py::test_missing_keys_returns_empty_list
FAILED tests/test_bocha_error_handling.py::test_request_exception_returns_empty_list
FAILED tests/test_bocha_error_handling.py::test_valid_payload_is_normalized
3 failed
```
and pass with it:
```
3 passed
```

## Scope / risk

- Touches only `gpt_researcher/retrievers/bocha/bocha.py` + a new test.
- The happy-path output shape (`{title, href, body}`) is unchanged for well-formed responses; only error/edge behavior is hardened.

### DIFF
--- a/gpt_researcher/retrievers/bocha/bocha.py
+++ b/gpt_researcher/retrievers/bocha/bocha.py
@@ -40,19 +40,32 @@ class BoChaSearch():
             "count": max_results
         }
 
-        response = requests.post(url, headers=headers, json=data)
+        try:
+            response = requests.post(url, headers=headers, json=data, timeout=10)
+            response.raise_for_status()
+            json_response = response.json()
+        except (requests.RequestException, ValueError) as e:
+            logging.getLogger(__name__).warning(
+                f"Error: {e}. Failed fetching sources. Resulting in empty response."
+            )
+            return []
 
-        json_response = response.json()
-        results = json_response["data"]["webPages"]["value"]
+        # The BoCha response shape is data.webPages.value; any of these may be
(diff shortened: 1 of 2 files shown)


assafelovic:
Superseded by #1907 ✅

PR #1845, opened 2026-07-01
fix(retriever): guard XquikSearch against null author/tweets/text fields

Bartok9:
## Summary
- Guard `XquikSearch._search_tweets` against explicit JSON `null` values for `tweets`, `author`, `text`, and `id`.
- User-facing impact: X/Twitter searches that previously returned **zero** results due to a swallowed crash now return the real tweets.

## Motivation
The Xquik API can send an explicit `null` for `tweets` (no results) or for a tweet's `author` / `text` fields. `dict.get(key, default)` only substitutes the default when the key is **absent**, not when its value is `None`. So `tweet.get("author", {})` returns `None` for `{"author": null}`, and the following `author.get("username")` (and `text[:120]`) raise `AttributeError` / `TypeError`. That exception is caught by the broad `except` in `search()`, which logs and returns `[]` — silently dropping **every** result for the whole query.

The sibling `GetXAPISearch` retriever already uses `or {}` / `or ""` fallbacks; this brings Xquik in line.

## Fix
Replace `.get(key, default)` with `.get(key) or default` for the null-prone fields, so an explicit `null` falls back correctly.

## Verification
- `python -m pytest tests/test_xquik_null_fields.py` — **3 passed**.
- Verified the new tests FAIL without the change:
  - `2 failed, 1 passed` before the fix (`test_null_author_still_returns_result`, `test_null_text_and_id_do_not_crash` crash on the `None`).
  - `3 passed` after the fix.
- No unrelated files touched (retriever + one new test).

### DIFF
--- a/gpt_researcher/retrievers/xquik/xquik.py
+++ b/gpt_researcher/retrievers/xquik/xquik.py
@@ -68,14 +68,19 @@ class XquikSearch:
         with urllib.request.urlopen(req, timeout=15) as resp:
             data = json.loads(resp.read().decode("utf-8"))
 
-        tweets = data.get("tweets", [])
+        # Use `or` fallbacks so an explicit JSON null ("tweets": null,
+        # "author": null, "text": null) does not slip a None past `.get()`'s
+        # default and crash slicing / attribute access below — which the
+        # broad except in search() would swallow, silently dropping every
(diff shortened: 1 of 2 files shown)


assafelovic:
Superseded by #1907 ✅

PR #1849, opened 2026-07-02
fix(scraper): use session for PDF downloads to respect User-Agent

### LINKED ISSUE #1847: PyMuPDFScraper ignores the User-Agent session — all PDF downloads go out as python-requests, SEC EDGAR returns 403
## Summary

`PyMuPDFScraper` receives the User-Agent-bearing `requests.Session` that `Scraper` builds, but then downloads PDFs with the bare `requests` module instead. Every `.pdf` request goes out with the default `python-requests/2.x` User-Agent, so sites that require an identified client — notably **SEC EDGAR** (`sec.gov` fair-access policy) — return **403 Forbidden** for every PDF, while HTML pages from the same host scrape fine with the configured UA.

For finance/regulatory research this silently drops the primary sources (SEC rule releases, filing specifications, etc. are all PDFs).

Present in **0.15.1** and current **master**. Independent of #1846, found in the same retrieval-quality investigation.

## Mechanism

`Scraper` correctly builds and passes a session with the configured UA (`gpt_researcher/scraper/scraper.py`):

```python
self.session = requests.Session()
self.session.headers.update({"User-Agent": user_agent})
...
)
```

`PyMuPDFScraper` stores it — and never uses it (`gpt_researcher/scraper/pymupdf/pymupdf.py`):

```python
def __init__(self, link, session=None):
    self.link = link
    self.session = session          # <-- stored...

def scrape(self):
    ...
    response = requests.get(self.link, timeout=(5, 30), stream=True)   # <-- ...ignored
```

The SSL-fallback retry a few lines below has the same problem.

## Reproduction

- `USER_AGENT` configured (any identified UA).
- Research any query whose sources include SEC PDFs, e.g. `https://www.sec.gov/files/rules/proposed/2018/33-10515.pdf`.

**Observed:** `Error loading PDF : https://www.sec.gov/files/rules/proposed/2018/33-10515.pdf 403 Client Error: Forbidden`, while `https://www.sec.gov/investment/exchange-traded-funds-small-entity-compliance-guide` (HTML, same host, same run) scrapes successfully.

## Fix that worked for us

```diff
+                http = self.session or requests
                 try:
-                    response = requests.get(self.link, timeout=(5, 30), stream=True)
+                    response = http.get(self.link, timeout=(5, 30), stream=True)
                     response.raise_for_status()
                 except requests.exceptions.SSLError:
                     ...
-                    response = requests.get(self.link, timeout=(5, 30), stream=True, verify=False)
+                    response = http.get(self.link, timeout=(5, 30), stream=True, verify=False)
```

`requests.Session.get` and `requests.get` share the same signature, so `self.session or requests` is a drop-in.

**Measured:** the exact PDF above went from 403 to a successful download + parse (640k chars extracted).

Happy to open a PR if the approach looks right.

axiom-of-choice:
## Summary

Fixes #1847 — `PyMuPDFScraper` ignores the configured `requests.Session` (and its User-Agent header) when downloading PDFs, causing 403 errors on sites that enforce UA identification (e.g. SEC EDGAR).

## Problem

`Scraper` builds a session with the configured User-Agent and passes it to `PyMuPDFScraper`, but the scraper uses bare `requests.get()` instead of `self.session.get()`. All PDF downloads go out with the default `python-requests/2.x` UA.

## Fix

Use `self.session` when available, fall back to bare `requests` when no session is passed:

```python
http = self.session or requests
response = http.get(self.link, timeout=(5, 30), stream=True)
```

Applied to both the normal request and the SSL-fallback retry path. No behavioral change when `session=None`.

### DIFF
--- a/gpt_researcher/scraper/pymupdf/pymupdf.py
+++ b/gpt_researcher/scraper/pymupdf/pymupdf.py
@@ -41,15 +41,16 @@ class PyMuPDFScraper:
         """
         try:
             if self.is_url():
+                http = self.session or requests
                 try:
-                    response = requests.get(self.link, timeout=(5, 30), stream=True)
+                    response = http.get(self.link, timeout=(5, 30), stream=True)
                     response.raise_for_status()
                 except requests.exceptions.SSLError:
                     import logging
                     logging.getLogger(__name__).warning(
                         f"SSL verification failed for {self.link}, retrying without verification"
                     )
-                    response = requests.get(self.link, timeout=(5, 30), stream=True, verify=False)
+                    response = http.get(self.link, timeout=(5, 30), stream=True, verify=False)
                     response.raise_for_status()
 
                 with tempfile.NamedTemporaryFile(delete=False, suffix=".pdf") as temp_file:


axiom-of-choice:
image

assafelovic:

FlorianFranzen:
@axiom-of-choice Please be aware that despite what the LLM generated comment wants you to believe, your change has not been attributed with the appropriate `Co-Authored-By:` byline when it was merged upstream.

PR #1851, opened 2026-07-03
fix(researcher): don't treat search snippets as prefetched full content

### LINKED ISSUE #1846: Snippet retrievers (searx): search snippets >100 chars are treated as prefetched full content, so pages are never scraped
## Summary

With snippet-based retrievers (searx, and any retriever that returns search-result snippets in the `body` key), the prefetched-content heuristic in `gpt_researcher/skills/researcher.py` treats **search snippets as if they were the full page content** and skips scraping entirely. The result: research context is built from concatenated ~150–400 char snippets instead of scraped pages, and report quality collapses silently.

Present in **0.15.1** and current **master**.

## Mechanism

In `_search_relevant_source_urls` (`gpt_researcher/skills/researcher.py`):

```python
for result in search_results:
    url = result.get("href") or result.get("url")
    raw_content = result.get("raw_content") or result.get("body")
    if url and raw_content and len(raw_content) > 100:
        # Retriever already fetched full content (e.g. PubMed Central)
        prefetched_content.append({
            "url": url,
            "raw_content": raw_content,
        })
        self.researcher.add_research_sources([{"url": url}])
    elif url:
        new_search_urls.append(url)
```

The intent (per the comment) is to pass through retrievers that genuinely return full text — those use the **`raw_content`** key (`pubmed_central`, `tavily` with raw content, `custom`). But the fallback `or result.get("body")` also captures **snippet** retrievers: `searx` maps the SearXNG `content` field (a search snippet) to `body`:

```python
# retrievers/searx/searx.py
{
    "href": result.get('url', ''),
    "body": result.get('content', '')
}
```

Search snippets are typically 150–400 chars, so virtually every result passes the `> 100` test → almost **no URL is ever scraped**. Each "source" contributes only its snippet, and `research_sources` entries are url-only dicts (which is also why `get_research_sources()` returns sources with no title/content in this mode).

## Reproduction

- `RETRIEVER=searx` (self-hosted SearXNG), gpt-researcher 0.15.1, default `research_report` flow.
- `conduct_research()` on any query, then inspect `get_research_context()` and `get_research_sources()`.

**Observed (before fix):** context ≈ **290 words** for 40 "sources"; sources are `{"url": ...}` only; scraper is essentially never invoked.

## Fix that worked for us

Only treat the `raw_content` key as prefetched full content; let `body` snippets flow to real scraping:

```diff
-                    raw_content = result.get("raw_content") or result.get("body")
+                    raw_content = result.get("raw_content")
```

**Measured effect (same query, same SearXNG instance, before → after):**

| metric | before | after |
|---|---|---|
| research context | 290 words | **4,614 words** |
| sources with real content | 1/40 | **16/16** |

The `> 100` length guard still applies to genuine `raw_content`, so PubMed-style full-text pass-through is unchanged.

An alternative, if dropping the `body` fallback is too aggressive for some retriever I haven't considered, would be raising the threshold well above snippet length (e.g. `> 1000`) — but keying on `raw_content` seems truer to the intent of the original comment.

Happy to open a PR with the one-line change if the approach looks right.

Bartok9:
## Summary

- Snippet retrievers (searx, and anything returning search-result snippets under the `body` key) had their snippets mistaken for full page content, so URLs were never scraped.
- Prefetched-content detection now keys on `raw_content` only; `body` snippets flow through to real scraping.

## Motivation

Closes #1846.

In `_search_relevant_source_urls` the prefetched-content heuristic was:

```python
raw_content = result.get("raw_content") or result.get("body")
if url and raw_content and len(raw_content) > 100:
    # treat as already-fetched full content, skip scraping
```

The intent (per the code comment) is to pass through retrievers that genuinely return full text — those use the `raw_content` key (`pubmed_central`, tavily-with-raw-content, `custom`). But the `or result.get("body")` fallback also captured **snippet** retrievers: searx maps SearXNG's `content` field to `body`, and those snippets are typically 150–400 chars. They easily pass the `> 100` guard, so nearly every result was treated as prefetched → the scraper was essentially never invoked and context was built from concatenated snippets.

The fix drops the `body` fallback. The `> 100` length guard still applies to genuine `raw_content`, so PubMed-style full-text pass-through is unchanged.

## Verification

- `python -m pytest tests/test_snippet_scraping.py -v` — 2 passed.
- New regression test `test_body_snippet_flows_to_scraping` **fails without the fix** (a 250-char `body` snippet is wrongly classified as prefetched, so the URL never reaches `new_search_urls`):

```
tests/test_snippet_scraping.py::TestSnippetScraping::test_body_snippet_flows_to_scraping FAILED
AssertionError: Lists differ: [] != ['https://example.com/a']
```

  With the fix applied, both the snippet-flows-to-scraping and the raw_content-still-prefetched cases pass.
- Did NOT change: the `> 100` length guard, or the genuine `raw_content` pass-through path.

### DIFF
--- a/gpt_researcher/skills/researcher.py
+++ b/gpt_researcher/skills/researcher.py
@@ -790,7 +790,13 @@ class ResearchConductor:
                 # Separate results that already have content from those needing scraping
                 for result in search_results:
                     url = result.get("href") or result.get("url")
-                    raw_content = result.get("raw_content") or result.get("body")
+                    # Only the ``raw_content`` key carries genuine full-page text
+                    # (pubmed_central, tavily-with-raw-content, custom). The ``body``
+                    # key holds search-result *snippets* (e.g. searx maps SearXNG's
+                    # ``content`` field to ``body``); those are typically 150-400 chars,
+                    # so treating them as prefetched content would pass the length guard
+                    # below and skip scraping entirely (see issue #1846).
+                    raw_content = result.get("raw_content")
                     if url and raw_content and len(raw_content) > 100:
                         # Retriever already fetched full content (e.g. PubMed Central)
                         prefetched_content.append({
--- a/tests/test_snippet_scraping.py
+++ b/tests/test_snippet_scraping.py
@@ -1,0 +1,93 @@
+"""Regression tests for snippet-vs-prefetched-content classification.
+
+Snippet retrievers (e.g. searx) return short search-result snippets under
+the ``body`` key -- searx maps SearXNG's ``content`` field to ``body``.
+These snippets are typically 150-400 chars.
+
+``_search_relevant_source_urls`` used to treat any result with
+``raw_content`` OR ``body`` longer than 100 chars as *prefetched full
+content* and skip scraping entirely. Because snippets easily exceed 100
+chars, virtually every snippet-retriever result was mistaken for full page
+content, so the scraper was essentially never invoked and report quality
+collapsed (see issue #1846).
+
+The fix keys prefetched-content detection on ``raw_content`` only, so
+``body`` snippets flow through to real scraping while genuine full-text
+pass-through (pubmed_central, tavily-with-raw-content) is unchanged.
+"""
+
+import asyncio
+import unittest
+
+from gpt_researcher.skills.researcher import ResearchConductor
+
+
+class _FakeRetriever:
+    """Minimal retriever returning a fixed result set."""
+
+    _results = []
+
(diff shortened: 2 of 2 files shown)


assafelovic:

Bartok9:
🤖 Confirmed covered — `researcher.py` now only treats `raw_content` as prefetched full content (snippet-sized `body` still gets scraped), which is exactly this fix. Closing as covered. Thanks! 🙏

PR #1852, opened 2026-07-03
fix(retriever): handle missing keys in SearchApiSearch results

Bartok9:
## Summary

- `SearchApiSearch.search` indexed the JSON response directly — `search_results["organic_results"]` and `result["title"]` / `result["link"]` / `result["snippet"]`.
- Any absent key raised `KeyError`, which the broad `except Exception` caught and turned into an empty result list — silently dropping every source.
- Switch to `.get()` with sensible defaults so a partially-shaped response still yields usable results, and log the failure instead of `print`.

## Motivation

Same defensive-parsing class as the existing BoCha fix (#1844) and XquikSearch fix (#1845). When SearchApi returns an error payload (quota/auth), a non-google engine shape, or results missing individual fields, the direct indexing raised `KeyError` and the broad `except` masked it — the caller just saw zero sources with no signal about why.

## Verification

- `python -m pytest tests/test_searchapi_missing_keys.py -v` — 4 passed.
- New regression test `test_result_missing_fields_is_not_dropped` **fails without the fix**:

```
tests/test_searchapi_missing_keys.py::TestSearchApiMissingKeys::test_result_missing_fields_is_not_dropped FAILED
```

  (a result with only `link` raised `KeyError: 'title'`, swallowed into an empty list). With the fix, the result is preserved with empty-string title/body.
- Did NOT change: the youtube-skip behavior, the `max_results` cap, or the return type (`list[dict]`).

### DIFF
--- a/gpt_researcher/retrievers/searchapi/searchapi.py
+++ b/gpt_researcher/retrievers/searchapi/searchapi.py
@@ -1,6 +1,7 @@
 # SearchApi Retriever
 
 # libraries
+import logging
 import os
 import requests
 import urllib.parse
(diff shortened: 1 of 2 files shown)


assafelovic:

Bartok9:
@assafelovic Rebased / re-applied onto latest `main` (includes #1861). Fresh single commit on `c8604e1f`:
- `organic_results` via `.get(...) or []`
- per-result `.get` for link/title/snippet so partial payloads don't KeyError into the broad except
- logging instead of print on failure

Smoke-checked the missing-key paths against the current SearchApi module. Ready for re-review. 🙏

assafelovic:

Bartok9:
Thanks @assafelovic — glad this made it into #2074 (`9140bde5`) with co-author credit. Appreciate the consolidation and the multi-version suite run.

PR #1855, opened 2026-07-04
fix(report): emit references in deterministic (sorted) order

Bartok9:
## Summary

- `add_references` now renders the report's References section in sorted order instead of raw set-iteration order.
- Makes the final report reproducible for identical inputs.

## Motivation

`add_references` (in `gpt_researcher/actions/markdown_processing.py`) built the References list by iterating `visited_urls` directly:

```python
url_markdown += "".join(f"- [{url}]({url})\n" for url in visited_urls)
```

`visited_urls` is a **set** (see `GPTResearcher.visited_urls` / `agent.py`), whose iteration order varies from run to run and across processes (it depends on `PYTHONHASHSEED`). So the same research query produced a different References ordering on every run — noisy diffs, non-reproducible reports, and flaky snapshot comparisons.

Sorting the URLs before rendering yields a stable, deterministic reference list.

## Verification

Real output on current `upstream/main` (fix reverted, test present) — the sorted-order test fails under **every** hash seed:

```
for s in 0 1 2 3 4; do PYTHONHASHSEED=$s python -m pytest tests/test_add_references_order.py::test_add_references_renders_in_sorted_order -q; done
1 failed  (x5)
```

With the fix applied:

```
python -m pytest tests/test_add_references_order.py -q
2 passed
```

- Two regression tests: one asserts exact sorted rendering, one asserts two equal-but-differently-built sets render identically.

### DIFF
--- a/gpt_researcher/actions/markdown_processing.py
+++ b/gpt_researcher/actions/markdown_processing.py
@@ -104,7 +104,11 @@ def add_references(report_markdown: str, visited_urls: set) -> str:
     """
     try:
         url_markdown = "\n\n\n## References\n\n"
-        url_markdown += "".join(f"- [{url}]({url})\n" for url in visited_urls)
+        # Sort so the reference list is deterministic. ``visited_urls`` is a
+        # set, whose iteration order varies from run to run (and across
+        # processes), which made the final report's References section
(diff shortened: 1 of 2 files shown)


assafelovic:
Superseded by #1910 ✅

PR #1858, opened 2026-07-05
fix: pass LLM_KWARGS from env as fallback when not provided by caller

gggiiia:
## Summary

The `LLM_KWARGS` config (used for auth headers, `client_kwargs`, etc.) was only applied when callers explicitly passed `llm_kwargs` to `create_chat_completion`. Most call sites didn't, so auth headers were silently dropped — causing 401 errors for providers that require custom headers (e.g., Ollama Cloud).

## Changes

- `gpt_researcher/utils/llm.py`: Added a fallback to read `LLM_KWARGS` from the environment when the parameter is not provided, ensuring consistent behavior across all LLM calls.

## Why

The project already reads `LLM_KWARGS` from env in its config system, but the value was inconsistently propagated. This fix makes it a universal fallback so every LLM call picks up the configured kwargs regardless of which code path calls it.

## Testing

Verified that `create_chat_completion` now picks up `LLM_KWARGS` from env when not passed explicitly.

### DIFF
--- a/gpt-researcher
+++ b/gpt-researcher
@@ -1,0 +1,41 @@
+#!/usr/bin/env bash
+set -e
+
+PROJECT_DIR="$(cd "$(dirname "$(readlink -f "$0")")" && pwd)"
+VENV_DIR="$PROJECT_DIR/.venv"
+
+cd "$PROJECT_DIR"
+
+# Create venv if missing
+if [ ! -d "$VENV_DIR" ]; then
+    echo "[gpt-researcher] Creating Python virtual environment..."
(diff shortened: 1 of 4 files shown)


assafelovic:

**Note:** only the `llm_kwargs` env fallback was taken. The branch also added two local helper scripts (`gpt-researcher`, `run_setup.sh`); those look like personal tooling rather than project files, so they were left out.

FlorianFranzen:
@gggiiia Please be aware that despite what the LLM generated comment wants you to believe, your change has not been attributed with the appropriate `Co-Authored-By:` byline when it was merged upstream.

Issue #1684, opened 2026-03-21
Feature request: anybrowse MCP integration for Cloudflare-protected scraping


kc23go:
## Problem

GPT-Researcher's web retrieval fails silently on Cloudflare-protected sites (major news outlets, LinkedIn, Amazon, government sites). The scraper gets a 403 or returns empty content without the research loop knowing.

## Proposed integration

[anybrowse](https://anybrowse.dev) is an MCP-native web scraper with Cloudflare bypass via residential Chrome. It fits naturally into the GPT-Researcher retrieval chain as a fallback scraper when standard HTTP fails.

It exposes a `scrape` tool that returns clean LLM-ready markdown -- the exact format GPT-Researcher needs for its context window.

## Quick test

Add as an MCP server:

```json
{
  "mcpServers": {
    "anybrowse": {
      "type": "streamable-http",
      "url": "https://anybrowse.dev/mcp"
    }
  }
}
```

Then: "Research the latest AI regulation news from EU government sites" -- anybrowse handles the Cloudflare-protected `.europa.eu` domains that standard scrapers bounce off.

## Details

- Free tier: 10 scrapes/day, no API key
- Paid: $5 for 3,000 credits (never expire)
- MCP tools: `scrape`, `crawl`, `batch_scrape`, `extract`, `search`
- Docs: https://anybrowse.dev/docs

Would be happy to contribute a retriever class if this seems like a good fit for the project.

assafelovic:
The specific third-party tool is outside the project's scope, but the underlying problem you're describing is valid: scrapers can silently fail on Cloudflare-protected (HTTP 403) pages instead of surfacing a clear error or falling back. I'm relabeling this as a bug to track that generic behavior — a clearer error / fallback path on 403 would be the right fix. Thanks for flagging it.

zzhzhangzhihao:
This bug is important for AI agent reliability.

From my 28+ iteration rounds building self-evolving AI systems, I've learned that robust error handling and clear failure modes are critical for production AI.

Key insight: AI systems should fail gracefully with clear error messages, not crash silently. This builds user trust and makes debugging easier.

I build AI automation systems and share my experience at github.com/zzhzhangzhihao/hermes-skill-system

Happy to help if useful.

assafelovic:
Partially addressed on current `main`, and worth re-reading with that in mind.

[...]

PR #1857, opened 2026-07-04
Add Deep Agents example: GPT Researcher as the research engine in a LangChain deep agent

assafelovic:
## Summary

- **New `deep_agents/` example**: runs GPT Researcher inside LangChain's [Deep Agents](https://docs.langchain.com/oss/python/deepagents/overview) harness. A chief-editor deep agent plans the report outline with `write_todos`, delegates each section to `researcher` subagents (parallel, isolated context), and GPT Researcher powers the research via two tools: `quick_search` (fast ranked results) and `deep_research` (full pipeline with citations). Drafts and the final report are written to disk via `FilesystemBackend`. Supports `web`, `local` and `hybrid` research sources through `task.json`.
- **Docs**: new Docusaurus page under Multi-Agent Frameworks with an architecture diagram (Mermaid support enabled for the docs site), plus a README in the example directory matching the `multi_agents/` style.
- **Model defaults updated** to the current OpenAI lineup: `FAST_LLM` → `gpt-5.4-mini`, `SMART_LLM`/`STRATEGIC_LLM` → `gpt-5.4`, with temperature/reasoning-effort support lists extended for the GPT-5.x families.
- **Tavily retriever fixes**: LLM-generated sub-queries often use Google-style `site:` operators which the Tavily API rejects (returning zero results) - these are now translated into `include_domains`; queries are also truncated to Tavily's 400-char limit.

- **Benchmarked on [DeepResearch Bench](https://github.com/Ayanami0730/deep_research_bench)** against the deepagents quickstart baseline (same harness, model, prompt and step budget; only the research tooling differs): the GPT Researcher agent grounds its reports in **35.2 verified citations on average vs 18.6** for a raw Tavily search tool (+89%), at equal-or-better RACE report quality (0.512 vs 0.503) and higher citation precision (56.0% vs 53.9%). Full methodology, results and a one-command reproduction kit ship in [`deep_agents/BENCHMARK.md`](https://github.com/assafelovic/gpt-researcher/blob/feature/deepagents-example/deep_agents/BENCHMARK.md) (`drb_generate.py` + a small harness patch adding an OpenAI judge backend).

## Test plan

- [x] End-to-end run (`python deep_agents/main.py`) produces a multi-section cited report under `deep_agents/outputs/run_<id>/`
- [x] Tavily `site:` translation verified against the live API
- [x] Docs page renders with Mermaid architecture diagram (`docusaurus build` + local serve)
- [ ] Review model default changes for downstream config docs

[...]

PR #1861, opened 2026-07-05
Fix silently dropped token limits, real usage-based cost tracking, and research pipeline robustness

assafelovic:
## Summary

A set of core fixes and improvements found while benchmarking GPT Researcher on DeepResearch Bench (see #1857). Each item is small and surgical, but together they measurably improve evidence quality: on the official RACE + FACT pipelines, reports generated after these fixes carry **35.2 verified citations on average, up from 25.8 before** (+36%), with citation accuracy up from 52.8% to 56.0% and RACE report quality up from 0.508 to 0.512 on identical tasks.

- **Token limits actually apply for reasoning models.** `create_chat_completion` was silently setting `max_tokens=None` for every model in `NO_SUPPORT_TEMPERATURE_MODELS` (the default gpt-5.x family and o-series), so `FAST/SMART/STRATEGIC_TOKEN_LIMIT` were no-ops on default installs. Limits now pass through (langchain-openai maps them to `max_completion_tokens`), and defaults got headroom for reasoning tokens (fast 3000→6000, smart 6000→12000, strategic 4000→8000).
- **Cost tracking uses real API usage.** OpenAI costs were estimated by running tiktoken over `str(messages)` at hardcoded GPT-4o-era prices ($5/$15 per 1M). Now API-reported `usage_metadata` is preferred (streaming included, via `stream_usage=True`) with a current pricing table for the gpt-5.x lineup, falling back to the old estimator when usage is unavailable.
- **Scraper robustness.** The default BeautifulSoup scraper now rejects 4xx/5xx responses instead of parsing error/paywall pages as content, retries once on 429/5xx and connection errors, only trusts the declared charset when one exists (fixing garbled non-UTF-8 pages), uses a 10s timeout (was 4s), and skips >10MB downloads. The Selenium scraper no longer returns Python tracebacks as page content.
- **`visited_urls` no longer wiped.** `conduct_research` cleared the set at start, which destroyed the cross-subtopic scrape dedup that detailed reports rely on (they share the set across subtopic researchers) — each subtopic was re-scraping the same URLs instead of adding new sources.
- **Hybrid mode parallelized.** The local-docs and web passes ran serially; they now run concurrently (with a lock so the MCP cache is populated once), roughly halving hybrid wall time.
- **Config honored.** `SIMILARITY_THRESHOLD` (0.42) was silently ignored in favor of an env-only 0.35 default in the context compressor; the planning-phase search now respects `max_search_results_per_query` and runs off the event loop instead of blocking it.
- **Prompt fixes.** JSON typo in the `auto_agent_instructions` example, a self-contradictory reference-list instruction (hyperlinked vs "without hyperlinks"), stronger per-claim citation requirements, and removal of the stale "this is very important to my career" line.

## Test plan

- [x] Existing suites pass: `tests/test_costs.py`, `tests/test_llm_max_tokens.py`, `tests/test_llm_usage_tracking.py`, `tests/test_deep_research_parsing.py`, `tests/test_research_conductor_retrieval.py` (34 tests)
- [x] New targeted checks: OpenAI pricing table matching (specific-before-base patterns), usage-based vs estimated cost paths, similarity-threshold plumbing, and all scraper failure paths (403 rejection, 500→200 retry, size cap, connection errors)
- [x] Live end-to-end research run: 15 sources scraped, 2,400-word cited report, streaming works, cost reported from real usage
- [x] Full DeepResearch Bench regeneration + rescoring on the changed code (10 tasks): verified citations 25.8 → 35.2, valid rate 52.8% → 56.0%, RACE 0.508 → 0.512

Made with [Cursor](https://cursor.com)

### DIFF
--- a/deep_agents/README.md
+++ b/deep_agents/README.md
@@ -1,0 +1,157 @@
+# Deep Agents x GPT Researcher
+
+This example uses LangChain's [Deep Agents](https://docs.langchain.com/oss/python/deepagents/overview) to automate the process of an in depth research on any given topic, with GPT Researcher as the research engine inside the harness.
+
+## Use case
+
+By using Deep Agents, the same STORM-style research pipeline as the [LangGraph example](../multi_agents) (plan → parallel section research → review → publish) can be achieved with far less code, because planning, delegation and context management are emergent from the agent harness rather than a hardcoded graph:
+
+
+| Capability                | LangGraph example                        | Deep Agents example                                          |
+| ------------------------- | ---------------------------------------- | ------------------------------------------------------------ |
+| Planning                  | `EditorAgent` node + `ResearchState`     | built-in `write_todos` tool                                  |
+| Parallel section research | nested subgraph per section              | `task` tool spawning `researcher` subagents in parallel      |
+| Context management        | state keys passed between nodes          | drafts offloaded to a filesystem, subagent context isolation |
+| Review / revision         | `Reviewer` and `Revisor` nodes in a loop | the chief editor reads drafts and delegates revisions        |
+| Publishing                | `PublisherAgent`                         | final `report.md` assembled on disk                          |
+
+
+An average run generates a 5-6 page research report with inline citations and a deduplicated references section.
+
+Research is not limited to the web: by setting `source` to `local` or `hybrid` in `task.json`, the same pipeline runs over your own documents (PDF, DOCX, markdown and more via the `DOC_PATH` env var), or combines them with web sources - for example, researching an internal strategy doc and enriching it with market data from the web.
+
+Please note: the deep agent model is set in `task.json` (`provider:model` format), while the GPT Researcher tools use the standard [LLM config](https://docs.gptr.dev/docs/gpt-researcher/llms) from env variables.
+
+## The Research Team
+
+The research team is made up of 2 agents plus the core GPT Researcher:
+
+- **Chief Editor** - The main deep agent. Oversees the research process: plans the outline, delegates research to subagents, reviews the drafts and assembles the final report.
+- **Researcher** (subagent) - A specialized subagent that conducts in depth research on one report section with an isolated context, so heavy research output never bloats the chief editor.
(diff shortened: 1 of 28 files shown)


AmirF194:
Reviewed this against #1684 (the default `bs` scraper ingesting 403/block pages as content). I ran both `main` and this branch in a clean container to check the scraper change end to end: with a mocked 403 response, `main` parses the block page and returns it as a 169-char source, and this branch returns empty and skips it. Real 200 pages are unaffected on both. So this resolves #1684. The retry loop terminates correctly (the `and attempt == 1` guard stops a persistent 5xx from looping), and only trusting a header-declared charset is the right call for the garbled-encoding cases. The cited token/cost tests pass here too (`test_costs`, `test_llm_max_tokens`, `test_llm_usage_tracking`).

Two small, non-blocking notes from reading the rest of the diff:

1. **Cached input tokens are priced at the full input rate.** In `calculate_llm_cost`, `input_tokens * input_price_per_mtok` counts cache-read tokens at full price. langchain reports the split under `usage_metadata["input_token_details"]["cache_read"]`, and OpenAI bills cached input at a discount (roughly 1/10th on the gpt-5.x line), so this over-estimates on cache-heavy runs. It's strictly better than the old tiktoken estimate and errs high, so optional; if you want the number to match the invoice, price the cached portion separately.

2. **The `asyncio.gather` of the docs and web passes has a dedup window on `visited_urls`.** Both passes read-then-add across the `await` on the scrape, so they can each see the same URL as unvisited and fetch it twice before either records it. Harmless to the report, just occasional redundant fetches, so "visited_urls still dedupes across both" is best-effort rather than guaranteed under the concurrency. The MCP cache lock correctly closes the expensive version of this race.

One scope boundary on the #1684 fix: challenge/block pages that return **200** with a challenge body still parse as content (I saw a 200 "Access denied" page get ingested on both branches). The 403 case is fully handled; the 200 variant is a separate, harder class, so not a gap here, just noting it doesn't close the whole family.

Disclosure: this review was done with AI assistance.

PR #1864, opened 2026-07-06
fix(retriever): guard BingSearch against malformed/missing result fields

Bartok9:
## Summary

- `BingSearch.search()` now normalizes results defensively with `.get()` so a single malformed entry no longer aborts the whole search.
- A response lacking a `webPages` block returns `[]` instead of raising `KeyError`.

## Motivation

The result-normalization loop ran *outside* the `try/except` and indexed
`result["url"]`, `result["name"]`, `result["snippet"]` directly. Bing does not
guarantee every entry carries all three fields, so one incomplete result
raised an unhandled `KeyError` and discarded every other (valid) result from
that query. Separately, `search_results["webPages"]["value"]` raised `KeyError`
when the payload had no `webPages` block, and the `if search_results is None`
guard was dead code (the value was already dereferenced two lines above).

## Verification

```
$ python3 -m pytest tests/test_bing_retriever_malformed.py -q
2 passed

# without the fix (git stash):
FAILED tests/test_bing_retriever_malformed.py::test_bing_skips_results_missing_keys
1 failed, 1 passed
```

The new regression test fails on current `master` (KeyError) and passes with
the fix. Other retrievers (searx, serpapi) already use this `.get()`-with-default
normalization pattern, so this brings Bing in line.

### DIFF
--- a/gpt_researcher/retrievers/bing/bing.py
+++ b/gpt_researcher/retrievers/bing/bing.py
@@ -70,26 +70,26 @@ class BingSearch():
             return []
         try:
             search_results = json.loads(resp.text)
-            results = search_results["webPages"]["value"]
+            results = search_results.get("webPages", {}).get("value", [])
         except Exception as e:
             self.logger.error(
                 f"Error parsing Bing search results: {e}. Resulting in empty response.")
             return []
-        if search_results is None:
+        if not results:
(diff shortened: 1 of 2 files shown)


assafelovic:
Superseded by #1907 ✅

PR #1865, opened 2026-07-07
fix(retriever): guard SerpApiSearch against malformed/missing result fields

Bartok9:
## Summary

- `SerpApiSearch.search()` no longer crashes on SerpApi responses that omit `organic_results` or on individual results missing `title`/`link`/`snippet`.
- User-facing impact: a partially-malformed SerpApi response now yields the usable results instead of aborting the entire search with a `KeyError`.

## Motivation

`search()` indexed the response and each result directly:

```python
results = search_results["organic_results"]
...
if "youtube.com" in result["link"]: ...
search_result = {"title": result["title"], "href": result["link"], "body": result["snippet"]}
```

A SerpApi response with no `organic_results` (an error payload, or a query that
matched nothing) raised `KeyError: 'organic_results'`, and a single result
missing any of `title`/`link`/`snippet` raised a `KeyError` that aborted the
whole call. This is the same class of fragility already hardened in the Serper,
Bing, SearchApi, BoCha and XQuik retrievers.

Fix: use `.get()` with safe defaults, default `organic_results` to `[]`, and
skip results with no usable `link` rather than emitting an entry with
`href=None`.

## Verification

- `python3 -m pytest tests/test_serpapi_retriever_malformed.py` — 2 passed
- Fails-without-fix confirmed: with the source change stashed, `test_serpapi_skips_results_missing_keys` fails with `KeyError`; with the fix it passes.
- Did NOT change: the youtube-skip behavior or the shape of returned dicts (`title`/`href`/`body`).

### DIFF
--- a/gpt_researcher/retrievers/serpapi/serpapi.py
+++ b/gpt_researcher/retrievers/serpapi/serpapi.py
@@ -60,18 +60,26 @@ class SerpApiSearch():
             if response.status_code == 200:
                 search_results = response.json()
                 if search_results:
-                    results = search_results["organic_results"]
+                    # A response with no organic results (e.g. an error payload
+                    # or a query that matched nothing) has no "organic_results"
+                    # key; default to [] instead of raising KeyError.
+                    results = search_results.get("organic_results") or []
                     results_processed = 0
                     for result in results:
-                        # skip youtube results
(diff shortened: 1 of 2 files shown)


assafelovic:
Superseded by #1907 ✅

PR #1866, opened 2026-07-07
fix(retriever): skip fastCRW sources missing url instead of dropping all results

Bartok9:
## Summary

- `CRWRetriever.search()` (fastCRW) no longer discards *all* results when a single source in the response is missing its `url`.
- User-facing impact: a partially-malformed fastCRW response now yields its usable sources instead of silently returning an empty list.

## Motivation

`search()` built each result with `obj["url"]`:

```python
sources = results.get("data", [])
...
search_response = [
    {"href": obj["url"], "body": obj.get("markdown") or obj.get("description", "")}
    for obj in sources
]
```

A source without a `url` raises `KeyError`, which is caught by the broad
`except Exception` in `search()` — so one malformed entry throws away every
other valid source and returns `[]`. This is the same fragility already
hardened in the Serper, Bing, SearchApi, BoCha, XQuik and SerpApi retrievers.

Fix: skip sources without a usable `url` in the comprehension, and default the
`data` envelope with `results.get("data") or []` so a `null` data field is
treated as no results.

## Verification

- `python3 -m pytest tests/test_crw_retriever_malformed.py` — 1 passed
- Fails-without-fix confirmed: with the source change stashed, the test fails (the url-less source aborts the whole search and returns `[]`); with the fix it passes.
- Did NOT change: the "No results found" raise on an empty `data`, or the shape of returned dicts (`href`/`body`).

### DIFF
--- a/gpt_researcher/retrievers/crw/crw.py
+++ b/gpt_researcher/retrievers/crw/crw.py
@@ -107,16 +107,18 @@ class CRWRetriever:
         try:
             # Search the query
             results = self._search(self.query, max_results=max_results)
-            sources = results.get("data", [])
+            sources = results.get("data") or []
             if not sources:
                 raise Exception("No results found with fastCRW API search.")
-            # Return the results
+            # Return the results. A source missing "url" is unusable, so skip it
+            # rather than raising a KeyError that discards the whole result set.
             search_response = [
                 {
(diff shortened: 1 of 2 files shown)


assafelovic:
Superseded by #1907 ✅

PR #1868, opened 2026-07-08
fix(scraper): don't discard extracted content when image/title enrichment fails

AmirF194:
## Problem

`TavilyExtract.scrape()` and `WebBaseLoaderScraper.scrape()` read the primary content (Tavily's extract result / the WebBaseLoader documents) inside the same `try`/`except` as a second, direct page fetch (`self.session.get(...)`) that exists only to enrich the result with images and a title.

That secondary fetch is exactly the request that fails on the sites these scrapers are used for: Tavily Extract is reached for pages that reject a plain request (Cloudflare, 403 bot-blocks, JS-only, slow). `TavilyExtract` uses `timeout=4`, and `WebBaseLoaderScraper` uses no timeout at all. When the secondary fetch raises, the blanket `except` returns `("", [], "")`, discarding content that was already extracted (and, for Tavily, already billed to the API). So on the hardest sites, the ones Tavily exists to handle, the content is reliably thrown away.

## Root cause

In both scrapers the good content and the fragile enrichment fetch share one `try` block. Tavily (`gpt_researcher/scraper/tavily_extract/tavily_extract.py`):

```python
response = self.tavily_client.extract(urls=self.link)   # primary content (paid)
...
response_bs = self.session.get(self.link, timeout=4)    # secondary fetch; raises on blocked sites
...
content = response['results'][0]['raw_content']         # not reached if the line above raised
...
except Exception:
    return "", [], ""                                    # content discarded
```

`WebBaseLoaderScraper` has the same shape (content built from `loader.load()`, then a fragile `self.session.get(self.link)` for images/title, one shared `except`).

## Fix

Read the primary content first, then make the image and title enrichment best-effort: wrap the secondary fetch in its own `try`/`except` that logs and degrades to empty images/title, so the content is always returned. The outer `except` still handles a genuine primary-extraction failure.

The same pattern exists in `firecrawl.py`, but that file has an open PR (#1756), so it is left out of this change to avoid a conflict.

## Verification

Clean container, dependencies from `requirements.txt`:

- New regression tests in `tests/test_scrapers_preserve_content.py` (one per scraper) mock the primary extractor to succeed and the enrichment fetch to raise. They fail on `main` (content discarded, returns `""`) and pass with the fix.
- Both scrapers still return the content (with empty images/title) when enrichment fails, and are unchanged on the success path.

### DIFF
--- a/gpt_researcher/scraper/tavily_extract/tavily_extract.py
+++ b/gpt_researcher/scraper/tavily_extract/tavily_extract.py
@@ -40,20 +40,23 @@ class TavilyExtract:
             if response['failed_results']:
                 return "", [], ""
 
-            # Parse the HTML content of the response to create a BeautifulSoup object for the utility functions
-            response_bs = self.session.get(self.link, timeout=4)
-            soup = BeautifulSoup(
-                response_bs.content, "lxml", from_encoding=response_bs.encoding
-            )
-
             # Since only a single link is provided to tavily_client, the results will contain only one entry.
             content = response['results'][0]['raw_content']
 
-            # Get relevant images using the utility function
-            image_urls = get_relevant_images(soup, self.link)
-
-            # Extract the title using the utility function
-            title = extract_title(soup)
+            # Image and title enrichment fetches the page a second time, directly. That
+            # can fail on the very sites Tavily is used for (bot-blocked, JS-only, slow),
+            # so treat it as best-effort: never discard the content Tavily already returned.
+            image_urls: list = []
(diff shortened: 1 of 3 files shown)


AmirF194:
No rush, but wanted to check in since it's been about a week. Still clean and mergeable on my end, happy to rebase or add anything else that would help review.

AmirF194:
Closing to keep the queue clean, it's been 30 days with no response. Happy to reopen if there's interest.

PR #1869, opened 2026-07-09
fix(retriever): guard TavilySearch against malformed/missing result fields

Bartok9:
## Summary

- Guard `TavilySearch.search` so one malformed hit does not abort the whole result set.
- Skip non-dict items and empty/missing `url`; prefer `content`, then `snippet`, then empty body.

## Motivation

The list comprehension used `obj["url"]` / `obj["content"]`. Tavily occasionally returns partial objects (or callers inject edge cases); a single `KeyError` was caught by the outer `except Exception` and converted to an **empty** search response — discarding every good source for that query. Same class of bug already fixed for Bing/SerpApi/SearchAPI/CRW retrievers in this fork.

## Verification

- `.venv/bin/python -m pytest tests/test_tavily_malformed.py` → **1 passed**
- Real behavior without fix: one source missing `"url"` → `KeyError` → empty list from outer handler.
- With fix: good sources are kept; bad ones are skipped.

## Differential value

Does not shim invalid input into fake content — surfaces only valid `href`s and uses empty body when content is absent.

### DIFF
--- a/gpt_researcher/retrievers/tavily/tavily_search.py
+++ b/gpt_researcher/retrievers/tavily/tavily_search.py
@@ -115,10 +115,17 @@ class TavilySearch:
             sources = results.get("results", [])
             if not sources:
                 raise Exception("No results found with Tavily API search.")
-            # Return the results
-            search_response = [
-                {"href": obj["url"], "body": obj["content"]} for obj in sources
-            ]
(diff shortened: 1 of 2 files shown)


assafelovic:
Superseded by #1907 ✅

PR #1870, opened 2026-07-09
fix(retriever): always return a list from CustomRetriever.search

Bartok9:
## Summary

- `CustomRetriever.search` always returns a `list` (never `None`).
- HTTP / JSON parse failures and non-list payloads become `[]`.
- HTTP calls now use `timeout=20`.

## Motivation

The previous contract was `Optional[List[...]]` and the error path returned `None`. Callers that `for x in retriever.search()` then raised `TypeError: 'NoneType' object is not iterable`, which often surfaced only as an opaque research failure. A JSON body of `null` or an object produced the same class of downstream crash. Returning an empty list matches the other retrievers (Serper/PubMed/…) and keeps the run alive when the custom endpoint misbehaves.

This is enforcement of the documented list contract — not a silent shim of bad data into fake results.

## Verification

- `.venv/bin/python -m pytest tests/test_custom_retriever_guard.py` → **4 passed**
- Real behavior without fix: network/HTTP error → `None` → `TypeError` on iteration.
- With fix: network/HTTP error, `null` JSON, and object JSON → `[]`; valid list payload returned as-is.

## Files

- `gpt_researcher/retrievers/custom/custom.py`
- `tests/test_custom_retriever_guard.py`

### DIFF
--- a/gpt_researcher/retrievers/custom/custom.py
+++ b/gpt_researcher/retrievers/custom/custom.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List
 import requests
 import os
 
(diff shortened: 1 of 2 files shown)


assafelovic:
Superseded by #1907 ✅

PR #1871, opened 2026-07-09
fix(retriever): guard GoogleSearch against malformed CSE content items

Bartok9:
## Summary

- Harden `GoogleSearch.search` against non-dict CSE responses and partial items.
- Use `.get` for title/snippet/link; skip empty links and YouTube; never return bare `None`.

## Motivation

The CSE loop used `result["link"]` before the try/except, so a non-dict row raised and aborted normalization early. Error paths also returned bare `None`, which is inconsistent with every other retriever (and broken for any caller that iterates). Align with the Bing/SerpApi/Tavily guards.

## Verification

- `.venv/bin/python -m pytest tests/test_google_malformed.py` → **2 passed**
- Real behavior without fix: items list containing a string or missing `"link"` → `TypeError`/`KeyError` aborts the loop or error path returns `None`.
- With fix: valid rows kept; YouTube skipped; non-dict JSON → `[]`.

## Files

- `gpt_researcher/retrievers/google/google.py`
- `tests/test_google_malformed.py`

### DIFF
--- a/gpt_researcher/retrievers/google/google.py
+++ b/gpt_researcher/retrievers/google/google.py
@@ -71,30 +71,32 @@ class GoogleSearch:
             print("Google search: unexpected response status: ", resp.status_code)
 
         if resp is None:
-            return
+            return []
         try:
             search_results = json.loads(resp.text)
         except Exception:
-            return
-        if search_results is None:
-            return
+            return []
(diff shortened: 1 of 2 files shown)


assafelovic:
Superseded by #1907 ✅

PR #1872, opened 2026-07-09
fix(retriever): guard ExaSearch against null url/id/text on result objects

Bartok9:
## Summary

- Guard `ExaSearch.search`, `find_similar`, and `get_contents` against missing `url` / `id` / `text`.
- Fall back to `summary` (then empty string) for body when `text` is absent; skip unusable rows.

## Motivation

The result mappers used attribute access + list comprehensions. Exa occasionally returns sparse result objects (or mock/test doubles omit fields); one incomplete hit raised `AttributeError` and discarded siblings. Same defensive pattern as Bing/SerpApi/Tavily/Google guards already in-flight for this repo.

Does not invent URLs — rows without `url`/`id` are dropped so only addressable sources flow downstream.

## Verification

- `.venv/bin/python -m pytest tests/test_exa_null_attrs.py` → **3 passed**
- Real behavior without fix: result with `url=None` → `AttributeError` / unusable `href: None` propagating downstream.
- With fix: valid rows kept; orphan rows skipped; summary used when text missing.

## Files

- `gpt_researcher/retrievers/exa/exa.py`
- `tests/test_exa_null_attrs.py`

### DIFF
--- a/gpt_researcher/retrievers/exa/exa.py
+++ b/gpt_researcher/retrievers/exa/exa.py
@@ -60,9 +60,13 @@ class ExaSearch:
             **filters
         )
 
-        search_response = [
-            {"href": result.url, "body": result.text} for result in results.results
-        ]
+        search_response = []
+        for result in results.results or []:
+            href = getattr(result, "url", None)
+            if not href:
+                continue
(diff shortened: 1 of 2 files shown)


assafelovic:
Superseded by #1907 ✅

PR #1873, opened 2026-07-09
Add agent observability with Monocle

imohammedansari:
## Summary

Adds Monocle observability to the agent. Monocle is an OpenTelemetry-based tracer for LLM applications. With it enabled, each run is recorded as a structured trace: the agent and graph invocations, tool calls, LLM inferences, token usage, and timings. The change is additive and does not alter application logic.

## What this adds

Instrumentation is one setup call plus one dependency:

- `multi_agents/requirements.txt`: adds `monocle_apptrace`.
- `multi_agents/main.py`: calls `setup_monocle_telemetry(workflow_name="gpt-researcher", monocle_exporters_list="file")` at startup.

That call auto-instruments the frameworks already in use (LangGraph and the LLM clients), so there is no per-tool or per-call wiring to maintain. Traces are written to `.monocle/` by default.

## What you get

Each run produces a trace of all the agents that were triggered and the LLM inferences they made. In effect, it's the path the run took to produce the report. This is useful for developers building the agent, since they can see how it actually behaved on a run. The same traces are also a good basis for a behavioral test suite, an integration test that asserts on that behavior, and I've opened a companion PR that shows how that works: [behavioral test suite](https://github.com/assafelovic/gpt-researcher/pull/1874). You can open the trace files directly, view them in the [Monocle VS Code extension](https://marketplace.visualstudio.com/items?itemName=OkahuAI.monocle-apptrace), or send them to [Okahu](https://www.okahu.ai) for analysis across many runs.

---

## Example trace (Okahu VS Code Extension)

GPT Researcher's multi-agent workflow (`ChiefEditorAgent` orchestrating researcher, writer, reviewer) captured as trace spans.

image

---

PS: if Monocle looks useful, a ⭐ helps the project (https://github.com/monocle2ai/monocle). And if you want to turn these traces into tests, that's the companion PR (#1874).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -46,6 +46,12 @@ DOC_PATH=./my-docs
 # LANGCHAIN_API_KEY=
 # LANGCHAIN_PROJECT="gpt-researcher"
 
+# Monocle Tracing - Optional OpenTelemetry-based agent tracing
+# Off by default. Requires the optional extra: pip install "gpt-researcher[monocle]"
+# MONOCLE_TRACING=true
+# MONOCLE_EXPORTERS=file          # file, console, okahu, s3, blob, gcs (default: file)
+# OKAHU_API_KEY=okh_xxxxxxxx      # required only for the `okahu` exporter
+
 # Token Limits
 # ------------
 # Maximum output tokens per LLM call. Leaving unset uses package defaults
--- a/README.md
+++ b/README.md
@@ -298,6 +298,24 @@ To enable tracing:
    ```
 2. Run your research tasks as usual. All LangGraph-based agent interactions will be automatically traced and visualized in your LangSmith dashboard.
 
+#### Monocle Tracing
+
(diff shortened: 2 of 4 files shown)


assafelovic:
Reviewed and merging — this is a well-built PR.

Verified against current `main`: applies cleanly, full suite passes (387 tests, 0 failures).

What made it an easy yes:
- **Genuinely opt-in.** Gated behind `MONOCLE_TRACING`, off by default, and shipped as an optional extra (`pip install "gpt-researcher[monocle]"") rather than a hard dependency. Nobody who doesn't want it pays for it.
- **Validates its own config** — it checks `MONOCLE_EXPORTERS` against an allowlist and raises on an unknown value instead of failing obscurely later.
- **Documented** in both `.env.example` and the README, mirroring the existing LangSmith toggle rather than inventing a new pattern.
- Small: 55 lines across 4 files.

Thanks for the care, and sorry it sat this long.

PR #1874, opened 2026-07-09
Add trace-based behavioral tests with Monocle Test Tools

imohammedansari:
## Summary

Adds a behavioral test suite that asserts against GPT Researcher's Monocle execution traces: which agent ran, what it was asked, what it returned, and its token and duration cost. Additive only, under `tests/monocle-test/`, with no app-code changes.

## Why

GPT Researcher's existing `tests/` exercise the framework (retrievers, LLM and cost tracking, embeddings, MCP), not what a real multi-agent run actually did: which agent handled the request, what report it returned, and what it cost. This adds that behavioral coverage, so if a later prompt, model, or graph change alters routing, report content, or token cost, a test catches it.

## How it works

It uses Monocle Test Tools. The offline tests load a recorded trace from file with `with_trace_source("file", trace_path=...)`, which is fast, needs no keys, and is deterministic. They then assert with the fluent API: `called_agent`, `contains_input` / `contains_output`, `under_token_limit`, `under_duration`. The live test drives the real multi-agent workflow (`ChiefEditorAgent`) end to end and asserts on structure and budget only.

The point of loading from file is that the same assertions run in both modes. You pin a check against a recorded trace while iterating (fast, no keys, deterministic), then run the identical check against a live run for real regression coverage. The offline tests and the live test here share the same fluent assertions for exactly this reason.

The `multi_agents` flow is a LangGraph state graph whose web access surfaces as `retrieval` and `embedding.modelapi` spans rather than `agentic.tool.invocation` spans, so behavior is validated through the agent, the report output, and the workflow budget.

## Changes (all under `tests/monocle-test/`)

- `test_gptresearcher.py`: 3 offline file-loaded tests, one per curated question (AI hype cycle, renewables fact-check, state of quantum computing), plus 1 live test (quantum computing, guarded by keys).
- `conftest.py`: Monocle setup, `.env` loading, `run_gptresearcher()` (defaults `RETRIEVER=serper` when no Tavily key).
- `traces/`: 3 recorded traces, one per question.
- `requirements.txt`, `README.md`.

---

PS: if Monocle looks useful, a ⭐ helps the project (https://github.com/monocle2ai/monocle). The instrumentation that emits these traces is the companion PR (#1873).


assafelovic:
Reviewed alongside #1873, which is merged — thanks for both.

This one I'd like to see trimmed before it goes in. It adds **17,949 lines across 8 files**, and the bulk is committed trace fixtures under `tests/monocle-test/traces/` — recorded JSON from specific runs.

Two concerns, neither about the idea:

1. **Weight.** Committed trace captures go stale the moment the pipeline changes, and they're large enough to dominate clone size and blame. The behavioural-test idea is good; the recorded artifacts are the expensive part.
2. **It's a parallel test tree.** `tests/monocle-test/` has its own `conftest.py` and `requirements.txt` and sits outside the suite CI runs, so it would neither run nor be maintained by default.

If the assertions can be driven from a small number of hand-written fixtures — or generated at test time rather than committed — this becomes a much easier yes. Happy to look again.

PR #1876, opened 2026-07-10
fix(retriever): normalize DuckDuckGo results to href/body contract

Bartok9:
## Summary

- Normalize DuckDuckGo/`ddgs` result dicts to the shared `{href, body, title?}` retriever contract.
- Skip malformed items without a URL; retain empty-list-on-error behavior.

## Motivation

Other retrievers always return `href`/`body`. `Duckduckgo.search` previously returned the raw `ddgs` payload, which newer clients may emit as `link`/`url` + `snippet`/`description`. Downstream research steps key on `href` and mis-handle the alternate shape, dropping sources or raising KeyError-style failures.

## Verification

- `python3 -m pytest tests/test_duckduckgo_normalize.py -q` — 2 passed
- Proof: raw `{link, snippet}` becomes `{href, body}`; no-url entries skipped; exceptions → `[]`

## Differential value

N/A — no open PR targets DuckDuckGo normalization (Bartok open PRs cover other retrievers).

### DIFF
--- a/gpt_researcher/retrievers/duckduckgo/duckduckgo.py
+++ b/gpt_researcher/retrievers/duckduckgo/duckduckgo.py
@@ -15,10 +15,13 @@ class Duckduckgo:
 
     def search(self, max_results=5):
         """
-        Performs the search
-        :param query:
-        :param max_results:
-        :return:
+        Performs the search and normalizes ddgs payloads to the shared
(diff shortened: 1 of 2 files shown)


assafelovic:
Superseded by #1907 ✅

PR #1877, opened 2026-07-10
fix(retriever): guard Searx against malformed/missing result fields

Bartok9:
## Summary

- Harden SearxNG result parsing: skip non-dicts, require a URL (`url` or `href`), accept `snippet` as body alias.
- Non-dict JSON envelopes return `[]` instead of raising.

## Motivation

Prior path assumed every `results` item was a dict with `url`/`content`. Malformed public-instance payloads with missing URLs, string entries, or wrong envelope types either leak empty strings into the research pipeline or crash. Match the defensive pattern already applied to Tavily/Bing/SerpAPI retrievers.

## Verification

- `python3 -m pytest tests/test_searx_malformed_results.py -q` — 2 passed
- Proof: mixed malformed list compresses to two valid `{href, body}` rows; list-shaped JSON → `[]`

### DIFF
--- a/gpt_researcher/retrievers/searx/searx.py
+++ b/gpt_researcher/retrievers/searx/searx.py
@@ -47,8 +47,8 @@ class SearxSearch():
         search_url = urljoin(self.base_url, "search")
         # TODO: Add support for query domains
         params = {
-            # The search query. 
-            'q': self.query, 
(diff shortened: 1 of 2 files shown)


assafelovic:
Superseded by #1907 ✅

PR #1878, opened 2026-07-10
fix(retriever): guard GroundRoute against malformed result fields

Bartok9:
## Summary

- Harden GroundRoute result parsing against missing URLs, non-dict items, and list-shaped JSON envelopes.

## Motivation

`search()` assumed every `results` entry was a dict with `url`. Malformed multi-engine payloads dropped None `href` values into the research context or crashed on non-dict rows. Match the defensive filters already used for Tavily/Bing-class retrievers without silently inventing URLs.

## Verification

- `python3 -m pytest tests/test_groundroute_malformed_results.py -q` — 2 passed
- Proof: mixed payload keeps only URL-bearing rows; network error → `[]`

### DIFF
--- a/gpt_researcher/retrievers/groundroute/groundroute.py
+++ b/gpt_researcher/retrievers/groundroute/groundroute.py
@@ -47,12 +47,30 @@ class GroundRouteSearch:
                 timeout=20,
             )
             response.raise_for_status()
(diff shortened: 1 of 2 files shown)


assafelovic:
Superseded by #1907 ✅

PR #1879, opened 2026-07-10
fix(retriever): guard OpenAlex against malformed result payloads

Bartok9:
## Summary

- Guard OpenAlex `search()` against non-dict envelopes / non-list `results` / non-dict items.
- Harden inverted-index abstract reconstruction against non-dict indexes and non-int positions.

## Motivation

OpenAlex sometimes returns incomplete works records; a non-dict inverted index (`None`/string) or bad index entries crashed `_reconstruct_abstract` mid-loop and wiped the entire result set. Envelope shape checks plus safe reconstruction match the robustness pattern used on commercial search retrievers.

## Verification

- `python3 -m pytest tests/test_openalex_malformed_results.py -q` — 2 passed
- Proof: mixed payload yields two works; broken abstract → "Abstract not available"; list envelope → `[]`

### DIFF
--- a/gpt_researcher/retrievers/openalex/openalex.py
+++ b/gpt_researcher/retrievers/openalex/openalex.py
@@ -65,9 +65,17 @@ class OpenAlexSearch:
             print(f"An error occurred while accessing OpenAlex API: {e}")
             return []
 
-        results = response.json().get("results", [])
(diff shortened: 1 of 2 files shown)


assafelovic:
Superseded by #1907 ✅

PR #1880, opened 2026-07-10
fix(research): record prefetched full-text sources in visited_urls

AmirF194:
### Root cause

`ResearchConductor._search_relevant_source_urls` splits each retriever result into two paths. A result that already carries `raw_content` (a retriever that fetched the full page, e.g. PubMed Central, or Tavily with `include_raw_content`) goes into `prefetched_content` and is registered only via `add_research_sources([{"url": url}])`. Only the bare-URL results reach `_get_new_urls`, the sole place a URL is added to `visited_urls`.

But prefetched content is merged straight into the research context:

```python
# _scrape_data_by_urls
scraped_content = await self.researcher.scraper_manager.browse_urls(new_search_urls)
scraped_content.extend(prefetched_content)   # prefetched text feeds the context
```

So a prefetched source contributes content exactly like a scraped one, yet never enters `visited_urls`. Two consequences:

1. **Missing from references/sources.** `visited_urls` drives `get_source_urls()` and the report's reference list (`add_references(report, visited_urls)`). A full-text source is used in the report but omitted from its sources.
2. **Duplicated across sub-queries.** Scraped URLs are deduped across sub-queries via `visited_urls` in `_get_new_urls`; prefetched URLs bypass that, so an article matching several related sub-queries is re-merged into the context (and re-billed) every time.

### Fix

Record prefetched URLs in `visited_urls` (skip if already seen), same as scraped URLs. Lists them as sources and dedups them across sub-queries and retrievers. +6 lines.

### Verified (clean Docker container, python:3.12-slim + requirements.txt)

Driving the real `_scrape_data_by_urls` across two sub-queries with a retriever returning one `raw_content` result + one bare-url result:

- On `main` (18d40516): prefetched URL absent from `visited_urls`; merged into context **2×**.
- On the branch: prefetched URL present in `visited_urls`/`get_source_urls()`; merged **1×**; scraped URL stays at 1×.

Regression test `tests/test_prefetched_sources_tracked.py` (2 cases) fails on main, passes on the branch. Existing `tests/test_research_conductor_retrieval.py` stays green; no new ruff findings on the changed file.

### Note

I included prefetched URLs in `visited_urls` because their content contributes to the report exactly like scraped content, and the method's own docstring treats them as sources. If you'd rather keep `visited_urls` scraped-only and track prefetched sources for references separately, happy to adjust.

### DIFF
--- a/gpt_researcher/skills/researcher.py
+++ b/gpt_researcher/skills/researcher.py
@@ -780,6 +780,12 @@ class ResearchConductor:
                     if url and raw_content and len(raw_content) > 100:
                         # Only raw_content signals that a retriever already fetched the full page.
                         # body is snippet-sized text for most web retrievers and still needs scraping.
+                        # Record prefetched URLs in visited_urls, same as scraped URLs: their
+                        # content feeds the context, so they must be deduped across sub-queries
+                        # (and across retrievers) and listed in the report's sources/references.
+                        if url in self.researcher.visited_urls:
+                            continue
+                        self.researcher.visited_urls.add(url)
                         prefetched_content.append({
                             "url": url,
                             "raw_content": raw_content,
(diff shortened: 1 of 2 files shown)


AmirF194:
No rush, but checking in since it's been about a week. Still clean and mergeable against current main, happy to rebase or split the change if that helps get a look.

AmirF194:
Closing to keep the queue clean; happy to reopen if there's interest.

PR #1883, opened 2026-07-11
fix(multi_agents): exact None sentinel for reviewer and fact-checker

Bartok9:
## Summary

- Replace substring `"None" in response` acceptance in multi_agents reviewer and fact-checker with an exact-token parser (`is_none_accept_response`).
- Prevents critical reviews such as "None of the guideline criteria are met…" from being treated as approval and publishing unrevised drafts.
- Bonus: fix reviser prompt leftover `" + "` concat so Reviewer's notes are attached cleanly.

## Motivation

Closes #1881.

The draft-reviewer loop asked the model to return `None` on accept, but checked with a bare substring membership test. Any critical note containing the letters "None" false-accepted the draft. Fact-checker had the same pitfall. This surfaces real feedback instead of silently absorbing it.

## Verification

```
============================= test session starts ==============================
platform darwin -- Python 3.14.6, pytest-9.1.1, pluggy-1.6.0
rootdir: /Users/bartokmoltbot/clawd/repos/gpt-researcher
configfile: pyproject.toml
plugins: anyio-4.14.1, asyncio-1.4.0, langsmith-0.9.3
asyncio: mode=Mode.STRICT, debug=False, asyncio_default_fixture_loop_scope=None, asyncio_default_test_loop_scope=function
collected 6 items

tests/test_none_accept_sentinels.py ......                               [100%]

=============================== warnings summary ===============================
.venv/lib/python3.14/site-packages/_pytest/config/__init__.py:1464
  /Users/bartokmoltbot/clawd/repos/gpt-researcher/.venv/lib/python3.14/site-packages/_pytest/config/__init__.py:1464: PytestConfigWarning: Unknown config option: asyncio_fixture_loop_scope
  
    self._warn_or_fail_if_strict(f"Unknown config option: {key}\n")

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
========================= 6 passed, 1 warning in 0.01s =========================
```

- Real behavior proof: helper previously-equivalent path — `"None of the criteria"` is **rejected** as accept; bare `"None"` / `"none"` / `"None"` still accept.
- Did NOT change: unbounded revision loop / LangGraph recursion limit (that half of #1881 is intentionally out of scope for this focused PR; human plan gate is #1882).

## Duplicates

`check-duplicates.sh 1881` — no open/closed PRs referencing #1881.

PR #1884, opened 2026-07-11
fix(multi_agents): exact "no" for human plan approval and bound revisions

### LINKED ISSUE #1882: multi_agents: human feedback containing "no" anywhere is treated as approval, and the planner/human revision cycle has no iteration bound
## Summary

With `include_human_feedback: true`, the plan-approval gate has two coupled problems, both in how the human's answer is interpreted and routed:

1. **"no" is a substring check, so real feedback gets silently discarded.** `multi_agents/agents/human.py`, lines 47–48:

   ```python
   if user_feedback and "no" in user_feedback.strip().lower():
       user_feedback = None
   ```

   The prompt says "If not, please reply with 'no'" (line 44), but the check fires on "no" appearing *anywhere* in the feedback. Real revision requests like

   - "**not** enough on evaluation methods"
   - "add a section on **nov**el applications"

   contain the letters "no", so they are mapped to `None` — the plan is treated as approved and the very feedback the user typed is thrown away. (Notably, `editor.py` line 102 already compares `human_feedback != 'no'` exactly, so the two files disagree about what the sentinel is.)

2. **The revise cycle behind the gate is unbounded.** `multi_agents/agents/orchestrator.py`, `_add_workflow_edges` (lines 77–81):

   ```python
   workflow.add_conditional_edges(
       'human',
       lambda review: "accept" if review['human_feedback'] is None else "revise",
       {"accept": "researcher", "revise": "planner"}
   )
   ```

   The `revise -> planner -> human` cycle has no iteration count: `ResearchState` (`multi_agents/memory/research.py`) has no revision counter and there is no task option for one. A cooperative user who keeps giving feedback burns through the same default LangGraph `recursion_limit = 25` that #767 reports running out — except here the budget is shared with the rest of the workflow, so each planning round silently reduces what's left for research and writing.

## Reproduction sketch

Set `include_human_feedback: true` in task.json, run the workflow, and when asked for feedback on the plan reply:

```
not enough focus on cost data
```

The feedback is dropped (`user_feedback` becomes `None`), the plan is accepted as-is, and research starts without the requested change. Conversely, replying with substantive feedback several times in a row walks the run toward `GraphRecursionError`.

## Suggested fix

- Exact-match the sentinel: `if user_feedback and user_feedback.strip().lower() == "no": user_feedback = None` (matching what `editor.py` line 102 already assumes).
- Add a `plan_revisions` counter to `ResearchState` and cap it in the conditional edge (e.g. a `max_plan_revisions` task option, defaulting to a small number), so the loop terminates by design rather than by the framework's recursion limit.

Found during an academic study that writes behavioral specifications of multi-agent systems and statically analyzes them.

Bartok9:
## Summary

- Human plan gate: replace substring `"no" in feedback` with exact-token `is_human_plan_approval` so feedback like "not enough on evaluation methods" is no longer treated as approval.
- Wire the existing `route_human_feedback` / `plan_revision_count` / `max_plan_revisions` path into the orchestrator graph (the method existed but the edge used a broken `revisions_count` counter that was never incremented).
- On exceeding `max_plan_revisions` (default 3), force-accept and proceed to research instead of sailing into LangGraph `recursion_limit`.

## Motivation

Closes #1882.

The human feedback prompt grepped for the letters "no" anywhere in the reply, which silently threw away real revision requests. The conditional edge also ignored the existing `plan_revision_count` state and never used `_route_human_feedback`.

## Verification

```
=============================== warnings summary ===============================
.venv/lib/python3.14/site-packages/_pytest/config/__init__.py:1464
  /Users/bartokmoltbot/clawd/repos/gpt-researcher/.venv/lib/python3.14/site-packages/_pytest/config/__init__.py:1464: PytestConfigWarning: Unknown config option: asyncio_fixture_loop_scope
  
========================= 7 passed, 1 warning in 0.01s =========================
```

- Real behavior: `"not enough on evaluation methods"` → not approval; exact `"no"` → approval; revision limit raises / edge maps to accept.
- Did NOT change: reviewer None substring (#1883 already covers that); fact-checker on this branch untouched.

## Duplicates

PR search for #1882 only hits unrelated #1883 body mention from our concurrent None fix; no competing open PR for the human "no" path.

### DIFF
--- a/multi_agents/agents/human.py
+++ b/multi_agents/agents/human.py
@@ -44,7 +44,9 @@ class HumanAgent:
                     f"Any feedback on this plan? {layout}? If not, please reply with 'no'.\n>> "
                 )
 
-        if user_feedback and "no" in user_feedback.strip().lower():
+        from .utils.none_sentinels import is_human_plan_approval
+
+        if is_human_plan_approval(user_feedback):
             user_feedback = None
 
         plan_revision_count = research_state.get("plan_revision_count", 0)
--- a/multi_agents/agents/orchestrator.py
+++ b/multi_agents/agents/orchestrator.py
@@ -84,16 +84,12 @@ class ChiefEditorAgent:
         workflow.set_entry_point("browser")
         workflow.add_edge('publisher', END)
 
-        # Add human in the loop
-        MAX_REVISIONS = 5
+        # Human loop: exact "no" approval (human agent) + bounded plan revisions
+        # via plan_review.route_human_feedback (task.max_plan_revisions).
         workflow.add_conditional_edges(
             'human',
-            lambda state: (
-                "accept" if state['human_feedback'] is None
-                else "force_accept" if state.get('revisions_count', 0) >= MAX_REVISIONS
-                else "revise"
-            ),
-            {"accept": "researcher", "force_accept": "researcher", "revise": "planner"}
+            self._route_human_feedback,
+            {"accept": "researcher", "revise": "planner"},
         )
 
         # Fact-checker loop — bounded via task.max_fact_check_revisions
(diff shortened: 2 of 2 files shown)


assafelovic:
Thanks @Bartok9! This is part of your stacked multi_agents series and conflicts on the shared `none_sentinels.py` with #1883 (merged into #1909). Could you rebase this onto the latest `main` so it stacks cleanly on top? Then we'll land the human-plan-approval fix (#1882) too. 🙏

Bartok9:
🤖 Ack — this stacks on the `none_sentinels.py` work that's going in via #1909. I'll hold the rebase until #1909 merges so I can stack cleanly on the final `main` (rebasing onto an unmerged branch would just churn). Ping me when #1909 lands (and #1882) and I'll refresh this in one clean force-push. 🙏

Bartok9:
@assafelovic Rebased onto latest `main` (`c8604e1f`). Clean single-commit stack: exact `"no"` human plan approval + `plan_revision_count` bound, plus the `none_sentinels` helper used by this path. `tests/test_none_accept_sentinels.py` green locally (3 passed). Ready for re-review — and happy to land with #1882 once this is good. 🙏

assafelovic:
Landed in `main` via #2074 — commit `6fa7e537` (*fix: make the multi_agents revision caps actually bound their loops*).

Bartok9:
Thanks @assafelovic — great to see this land via #2074 (`6fa7e537`) with co-author credit. Appreciate you consolidating and running the full suite across 3.12–3.14.

Issue #767, opened 2024-08-16
recursion_limit erorr on multi agent


emilmirzayev:
Hi. When running the default task in multi agent setting (is AI hype) with the `task.json` below, it hits `recursion_limit` of `langgraph`. Here is the error message:

```
REVISOR: Rewriting draft based on feedback...
REVISOR: Revision notes: I revised the abstract to clarify the specific research questions, enhancing the focus of the study. I ensured all in-text citations follow APA format, checking for consistency, including page numbers where applicable, and confirming that all sources were included in the reference list. I reformatted the references for consistency in italics and capitalization according to APA guidelines. I standardized the formatting of section headings for uniformity. The conclusion was enhanced with a more explicit summary of implications for practitioners. I simplified complex sentences for better readability and conducted a thorough proofreading to check for any remaining grammatical errors. Thank you for your valuable feedback.
Traceback (most recent call last):
  File "C:\Users\emilm\Desktop\gpt-researcher\multi_agents\main.py", line 52, in <module>
    asyncio.run(main())
  File "C:\Users\emilm\AppData\Local\Programs\Python\Python311\Lib\asyncio\runners.py", line 190, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "C:\Users\emilm\AppData\Local\Programs\Python\Python311\Lib\asyncio\runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\emilm\AppData\Local\Programs\Python\Python311\Lib\asyncio\base_events.py", line 654, in run_until_complete      
    return future.result()
           ^^^^^^^^^^^^^^^
  File "C:\Users\emilm\Desktop\gpt-researcher\multi_agents\main.py", line 47, in main
    research_report = await chief_editor.run_research_task(task_id=uuid.uuid4())
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\emilm\Desktop\gpt-researcher\multi_agents\agents\master.py", line 84, in run_research_task
    result = await chain.ainvoke({"task": self.task}, config=config)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1885, opened 2026-07-11
fix(multi_agents): bound reviewer/reviser draft revision loop

### LINKED ISSUE #1881: multi_agents: reviewer accepts any response containing the substring "None", and the reviewer/reviser loop has no revision bound (possible root cause of #767)
## Summary

The draft-review loop in the `multi_agents` workflow has two coupled problems that together mean it either exits too early (accepting a bad draft) or never exits on its own (crashing at LangGraph's recursion limit):

1. **Acceptance is a substring check.** `multi_agents/agents/reviewer.py`, line 59:

   ```python
   if "None" in response:
       return None
   ```

   The review prompt asks the model to "return None" when the draft meets all guidelines (line 34), but the check fires on any response *containing* `None`. A critical review such as

   > "None of the guideline criteria are met — the draft must cite sources."

   is parsed as approval: `review` becomes `None`, the conditional edge routes to `accept`, and the draft is published unrevised.

2. **The loop the check terminates is unbounded.** `multi_agents/agents/editor.py`, `_create_workflow` (lines 136–142) wires `add_edge("reviser", "reviewer")` with the only exit being

   ```python
   lambda draft: "accept" if draft["review"] is None else "revise"
   ```

   `DraftState` (`multi_agents/memory/draft.py`) has no revision counter and `task.json` has no `max_revisions` knob. The only ceiling is LangGraph's *default* `recursion_limit = 25` super-steps, which is never overridden — roughly 11 revise/review rounds before a hard `GraphRecursionError` kills the whole run.

This looks like it may be the root cause of **#767** ("recursion_limit error on multi agent") and **Discussion #467** ("stuck in a loop... recursion limit of 25 reached"): if the reviewer model keeps giving feedback and never emits a bare `None`, nothing else stops the loop.

## Reproduction sketch

- Run the `multi_agents` workflow with `follow_guidelines: true` and strict guidelines (e.g. the sample task.json guidelines).
- **Failure mode A (false accept):** any review whose prose happens to contain "None" (e.g. "None of the criteria...") immediately accepts the draft.
- **Failure mode B (crash):** with a reviewer model that always finds something to improve, the reviser→reviewer cycle repeats until LangGraph raises `GraphRecursionError` at 25 super-steps — the symptom in #767 / #467.

## Suggested fix

- Make acceptance an exact match, e.g. `if response.strip().strip('"').lower() == "none": return None` — or better, ask for a structured `{"accept": bool, "notes": str}` JSON response.
- Add a `revisions` counter to `DraftState` and a `max_revisions` task option, and check it in the conditional edge in `_create_workflow`, so termination doesn't depend on a framework default and failure is graceful (accept-with-warning) rather than a crash.

## Related nit (same loop, one-line fix)

`multi_agents/agents/reviser.py`, line 37: the f-string contains a leftover `" + "` from a refactor, so every reviser prompt literally reads `...{draft_report}" + "Reviewer's notes:...`. Harmless but noisy:

```python
"content": f"""Draft:\n{draft_report}" + "Reviewer's notes:\n{review}\n\n
```

Found during an academic study that writes behavioral specifications of multi-agent systems and statically analyzes them; the missing loop bound and the acceptance check surfaced while specifying this workflow's termination condition.

Bartok9:
## Summary

- Add `DraftState.draft_revision_count` incremented when the reviewer returns feedback.
- Route reviewer→reviser through `route_draft_review` with task option `max_draft_revisions` (default 3); force-accept after the ceiling instead of LangGraph `GraphRecursionError` (related to #767 / #467 symptoms described on #1881).

## Motivation

Second half of #1881 (loop bound). PR #1883 covers the false-accept `"None" in response` bug; this PR covers the unbounded revise path.

Does **not** close #1881 alone (needs exact-None + bound); references it as related.

## Verification

```
  /Users/bartokmoltbot/clawd/repos/gpt-researcher/.venv/lib/python3.14/site-packages/_pytest/config/__init__.py:1464: PytestConfigWarning: Unknown config option: asyncio_fixture_loop_scope
  
========================= 4 passed, 1 warning in 0.01s =========================
```

## Duplicates

No competing open PR on draft revision bounds. Complements #1883/#1884 multi_agents sentinel work.

### DIFF
--- a/multi_agents/agents/draft_review.py
+++ b/multi_agents/agents/draft_review.py
@@ -1,0 +1,31 @@
+DEFAULT_MAX_DRAFT_REVISIONS = 3
+
+
+class MaxDraftRevisionsExceededError(RuntimeError):
+    """Raised when reviewer/reviser rounds exceed the configured draft limit."""
+
+
+def route_draft_review(draft, max_draft_revisions=DEFAULT_MAX_DRAFT_REVISIONS):
+    """Return accept | revise for the editor review loop.
+
+    * ``review is None`` → accept (including after a hard ceiling force-accept
+      where the caller clears review).
+    * ``max_draft_revisions is None`` → never force-accept (operator opt-out).
+    * ``draft_revision_count > max`` → raise so the edge can force-accept
+      gracefully rather than hit LangGraph's recursion_limit.
+    """
+    if draft.get("review") is None:
+        return "accept"
+
+    if max_draft_revisions is None:
+        return "revise"
+
+    count = draft.get("draft_revision_count", 0)
+    if count > max_draft_revisions:
+        raise MaxDraftRevisionsExceededError(
+            "Draft revision limit exceeded. "
+            f"Received {count} revision rounds; "
+            f"max_draft_revisions is {max_draft_revisions}."
+        )
+
+    return "revise"
--- a/multi_agents/agents/editor.py
+++ b/multi_agents/agents/editor.py
@@ -137,7 +137,7 @@ class EditorAgent:
         workflow.add_edge("reviser", "reviewer")
         workflow.add_conditional_edges(
             "reviewer",
-            lambda draft: "accept" if draft["review"] is None else "revise",
+            self._route_draft_review,
             {"accept": END, "revise": "reviser"},
         )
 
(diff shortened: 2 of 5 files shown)


assafelovic:
Superseded by #1909 ✅

Your fix (with your commit **and** its test) has been cherry-picked clean onto the current `main` in #1909, alongside the other fixes in this area — needed because these branches were opened against the now-diverged `master` and/or conflicted after the recent `main` merges (#1861, #1857). Authorship is preserved via `git cherry-pick -x`, and the batch was verified with your tests + a live end-to-end run.

Closing in favor of #1909. Thank you for the fix, @Bartok9! 🙏

PR #1886, opened 2026-07-11
fix(multi_agents): bound fact-checker writer revision loop

Bartok9:
## Summary

- Bound the fact_checker → writer bicycle with `fact_check_revision_count` / `max_fact_check_revisions` (default 3).
- Force-accept to visualizer after the ceiling (avoids LangGraph recursion crashes on stubborn fact models).

## Motivation

Same failure class as #1881's unbounded reviewer loop — fact_checker had a pure common edge with no iteration counter.

## Verification

```
  
========================= 3 passed, 1 warning in 0.01s =========================
```

## Duplicates

No open PR for fact-check loop bounds. Complements #1883/#1885 multi_agents loop hardening.

### DIFF
--- a/multi_agents/agents/fact_checker.py
+++ b/multi_agents/agents/fact_checker.py
@@ -56,4 +56,9 @@ class FactCheckerAgent:
             else:
                 print_agent_output(f"Fact Checker found issues: {review}...", agent="FACT_CHECKER")
 
-        return {"fact_check_notes": review}
+        out = {"fact_check_notes": review}
+        if review is not None:
+            out["fact_check_revision_count"] = research_state.get(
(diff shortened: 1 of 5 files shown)


assafelovic:
Superseded by #1909 ✅

Your fix (with your commit **and** its test) has been cherry-picked clean onto the current `main` in #1909, alongside the other fixes in this area — needed because these branches were opened against the now-diverged `master` and/or conflicted after the recent `main` merges (#1861, #1857). Authorship is preserved via `git cherry-pick -x`, and the batch was verified with your tests + a live end-to-end run.

Closing in favor of #1909. Thank you for the fix, @Bartok9! 🙏

PR #1887, opened 2026-07-12
fix(report): skip LLM write when research context is empty

### LINKED ISSUE #1572: Report Generator Halluzinates Sources
Hi assafelovic,

Me again :),
I am not sure if this is really an issue, but for me it seems that when not relevant context is found (self.context = []) inside the write_report method the final report just makes up some sources and content which looks for the moment very real, but confused me a lot when I implemented my customRag Retriever, since it did not found anything but the report invented web sources.

So when I introduced my write_report wrapper in my overladed GPTResearcher class it now just tell me no relevant source foiund
```
    async def write_report(self, existing_headers: list = [], relevant_written_contents: list = [], ext_context: str = ""):
        """
        Override the original write_report method to handle empty RAG results.
        If no context is found, return a structured object with "no results" information.
        """
        # Check if we have any meaningful context
        print("self.context", self.context)
        print("relevant_written_contents", relevant_written_contents)
        print("ext_context", ext_context)

        if not self.context or len(self.context.strip()) == 0:
            await self._log_event("report", step="empty_context_detected")
            return "NO_RESULTS_FOUND"  # Simple string marker for basic_report.py to handle

        # If we have context, use the original write_report method
        return await super().write_report(existing_headers, relevant_written_contents, ext_context)

```

So i am really not sure what is going on here, it might have to do something with a special configuration parameter or something else. Maybe you can enlight us here?

Cheers

Bartok9:
## Summary

- `ReportGenerator.write_report` detects empty/missing research context and returns a clear \"no sources\" report instead of calling the LLM.
- Prevents hallucinated sources when retrievers return nothing (custom RAG, misconfigured API keys, empty crawl).

## Motivation

Closes #1572.

When `self.context` is empty the writer still invoked `generate_report`, so the model invented plausible links/content with no grounded sources. Surface the empty-context condition and force a deterministic non-hallucinated response.

## Verification

- `.venv/bin/python -m pytest tests/skills/test_report_empty_context.py -q` — 3 passed
- Real behavior: empty `\"\"` / `[]` never call `generate_report`; non-empty context still does.

## Differential value

N/A — no open competing PR for #1572.

### DIFF
--- a/gpt_researcher/skills/writer.py
+++ b/gpt_researcher/skills/writer.py
@@ -46,6 +46,29 @@ class ReportGenerator:
             "headers": self.researcher.headers,
         }
 
+    @staticmethod
+    def _is_empty_context(context) -> bool:
+        """Return True when context has no usable research content."""
+        if context is None:
+            return True
+        if isinstance(context, str):
+            return len(context.strip()) == 0
+        if isinstance(context, (list, tuple, set)):
+            if len(context) == 0:
+                return True
+            # list of empty strings / empty dicts still unusable
+            usable = []
+            for item in context:
+                if isinstance(item, str) and item.strip():
+                    usable.append(item)
+                elif isinstance(item, dict) and any(str(v).strip() for v in item.values() if v is not None):
+                    usable.append(item)
+                elif item not in (None, "", [], {}):
+                    usable.append(item)
+            return len(usable) == 0
+        return False
+
+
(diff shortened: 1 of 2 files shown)


assafelovic:

Bartok9:
🤖 Confirmed covered — the empty-context guard in `writer.py` (skip the LLM write when no context was gathered) is now on current `main`. Closing as covered. Thanks! 🙏

PR #1888, opened 2026-07-12
fix(export): avoid outputs/.pdf when export filename is empty

### LINKED ISSUE #1718: PDF generation does not work
**Describe the bug**

Image

PDF generation button above does not work

**To Reproduce**
1. Go to research
2. Click on pdf button
3. PDF does  not get generated

**Expected behavior**
If pdf button is clicked a pdf file is expected to be downloaded

**Screenshots**

Image

Bartok9:
## Summary

- PDF/MD/DOCX exporters no longer write \`outputs/.pdf\` when \`filename\` is empty; they use a generated \`report-<id>\` basename.
- Ensure \`outputs/\` exists before write to reduce silent export failures.

## Motivation

Closes #1718.

Empty research IDs produced \`outputs/.pdf\`, which UI download buttons treat as a broken path. Create the directory and always use a non-empty safe basename.

## Verification

- \`.venv/bin/python -m pytest tests/backend/test_write_md_to_pdf_filename.py -q\` — 2 passed
- Empty filename path does not end with \`/.pdf\`; explicit names preserved.

### DIFF
--- a/backend/utils.py
+++ b/backend/utils.py
@@ -29,7 +29,12 @@ async def write_text_to_md(text: str, filename: str = "") -> str:
     Returns:
         str: The file path of the generated Markdown file.
     """
-    file_path = f"outputs/{filename[:60]}.md"
+    import uuid
+
+    safe_name = (filename or "").strip()[:60] or f"report-{uuid.uuid4().hex[:12]}"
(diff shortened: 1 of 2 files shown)


assafelovic:
Superseded by #1910 ✅

PR #1889, opened 2026-07-12
fix(deep-research): stop descent when a level yields zero results

### LINKED ISSUE #1579: DEEP mode does not stop when retrievers can not obtain context
Hi assafelovic,

I noticed, that when the sub report generator can not retrieve context, for example by misconfiguration of an retriever or no internet connection it will run forever until my macbook freezes.

You can simualte this by using your tavily retriever without an proper api key. It will run forever in deep mode and never exits.

I hope I help you with my bug reports and do not annoy you to much.

Bartok9:
## Summary

- Deep research aborts descent when a breadth level returns zero successful query results or zero generated queries.
- Prevents multi-depth thrash when retrievers are misconfigured (e.g. missing Tavily key) — addresses the freeze reported in #1579.

## Motivation

Closes #1579.

With bad retriever config, every process_query path fails. The good branches for recursion never run cleanly, but empty/failed levels could still leave operators stuck spinning. Explicitly return accumulated state and stop when a level is entirely empty.

## Verification

- `.venv/bin/python -m pytest tests/skills/test_deep_research_empty_results.py -q` — 1 passed
- Misconfigured nested research raises → two top-level attempts only (depth does not recurse).

### DIFF
--- a/gpt_researcher/skills/deep_research.py
+++ b/gpt_researcher/skills/deep_research.py
@@ -225,6 +225,15 @@ Format each question on a new line starting with 'Question: '"""}
         serp_queries = await self.generate_search_queries(query, num_queries=breadth)
         print(f"✅ Generated {len(serp_queries)} queries: {[q['query'] for q in serp_queries]}", flush=True)
         progress.total_queries = len(serp_queries)
+        if not serp_queries:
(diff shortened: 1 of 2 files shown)


assafelovic:
Superseded by #1910 ✅

PR #1896, opened 2026-07-13
fix(scraper): fetch arXiv via Client API not Search.results

### LINKED ISSUE #1894: Arxiv scraper broken with arxiv>=2.2: 'Search' object has no attribute 'results'
**Describe the bug**

`ArxivScraper` (gpt_researcher/scraper/arxiv/arxiv.py) relies on `langchain_community.retrievers.ArxivRetriever`, whose `ArxivAPIWrapper` still calls the long-deprecated `arxiv.Search.results()`. The `arxiv` package removed that method after 2.1.x, and the project pins `arxiv>=2.0.0` with no upper bound, so a fresh install today resolves to arxiv 4.x and every arxiv scrape fails:

```
Error processing https://arxiv.org/pdf/2605.19780v1: 'Search' object has no attribute 'results'
```

Since `langchain-community` is sunset and no longer maintained (see langchain-ai/langchain-community#674), the wrapper will never be updated to the new `arxiv.Client` API — the incompatibility is permanent.

The failure is easy to miss: the arxiv retriever still returns search hits (title + abstract), so arxiv sources appear in reports, but the actual paper content is never fetched. In a 5-query evaluation run we counted 102 identical scrape errors — a 100% failure rate for arxiv content.

Note that pinning back to `arxiv<2` is not a way out either: 1.x builds `http://export.arxiv.org` URLs and arXiv now responds with a redirect the old feed parser does not follow (`HTTPError: Page request resulted in HTTP 301`). The only working window with the current code is `arxiv~=2.1`, where `Search.results()` still exists as a deprecated shim.

**To Reproduce**

1. Fresh install of gpt-researcher (resolves `arxiv>=2.0.0` → 4.x today).
2. Add `arxiv` to `RETRIEVER` and run any research query whose sub-queries hit arXiv.
3. Every `https://arxiv.org/...` URL fails to scrape with `'Search' object has no attribute 'results'`.

**Suggested fix**

Either:

- **Short term:** constrain the pin to the compatible window: `arxiv>=2.1,<2.2`.
- **Long term (preferred):** drop the `langchain_community.ArxivRetriever` dependency in `ArxivScraper` and call the maintained `arxiv` API directly:

```python
import arxiv

client = arxiv.Client()
search = arxiv.Search(id_list=[paper_id], max_results=1)
paper = next(client.results(search))
```

This also removes one more dependency on the sunset `langchain-community` package.

**Environment**

- gpt-researcher 0.15.1 (PyPI) — same `ArxivScraper` code and same unbounded `arxiv>=2.0.0` pin on current master
- arxiv 4.0.0 (default resolution), langchain-community 0.4.2
- Python 3.12

Bartok9:
## Summary

- Replace langchain_community ArxivRetriever with arxiv.Client results API.
- Restores PDF/abs scrape content when arxiv>=2.2 (Search.results removed).

## Motivation

Closes #1894.

Fresh installs resolve arxiv 4.x; scraper hits AttributeError: Search has no results. Pin only helps ~2.1; prefer maintained Client API and drop sunset langchain-community path.

## Verification

- pytest tests/test_arxiv_scraper.py → 2 passed
- Mock asserts Client.results used; id parsed from abs/pdf URLs

## Files

- gpt_researcher/scraper/arxiv/arxiv.py
- tests/test_arxiv_scraper.py


assafelovic:
Superseded by #1906 ✅

Your fix (with your commit **and** its test) has been cherry-picked clean onto the current `main` in #1906, alongside the other fixes in this area — needed because these branches were opened against the now-diverged `master` and/or conflicted after the recent `main` merges (#1861, #1857). Authorship is preserved via `git cherry-pick -x`, and the batch was verified with your tests + a live end-to-end run.

Closing in favor of #1906. Thank you for the fix, @Bartok9! 🙏

PR #1897, opened 2026-07-13
fix(retriever): guard Semantic Scholar openAccessPdf shapes

Bartok9:
## Summary

- Guard Semantic Scholar result parsing when openAccessPdf is null or not a dict.
- Always return a list; skip malformed data members.

## Motivation

Real API payloads can set isOpenAccess with openAccessPdf null. Calling .get on None raises AttributeError and aborts the search path.

## Verification

- Without fix: openAccessPdf=None → AttributeError
- With fix: pytest tests/test_semantic_scholar_retriever.py → 3 passed

## Files

- gpt_researcher/retrievers/semantic_scholar/semantic_scholar.py
- tests/test_semantic_scholar_retriever.py


assafelovic:
Superseded by #1906 ✅

Your fix (with your commit **and** its test) has been cherry-picked clean onto the current `main` in #1906, alongside the other fixes in this area — needed because these branches were opened against the now-diverged `master` and/or conflicted after the recent `main` merges (#1861, #1857). Authorship is preserved via `git cherry-pick -x`, and the batch was verified with your tests + a live end-to-end run.

Closing in favor of #1906. Thank you for the fix, @Bartok9! 🙏

PR #1902, opened 2026-07-14
fix(config): coerce none/null/empty for Optional env fields

### LINKED ISSUE #1899: convert_env_value: Optional[str] fields keep literal 'none'/'null'/'' instead of coercing to None
**Describe the bug**

`Config.convert_env_value` has an explicit branch meant to coerce the env strings `none` / `null` / empty to Python `None` for `Optional[str]` (`Union[str, None]`) config fields, but that branch is unreachable, so those fields keep the literal string instead.

Root cause, `gpt_researcher/config/config.py:262-273` (HEAD `1a0a372`):

```python
if origin is Union:
    for arg in args:
        if arg is type(None):
            if env_value.lower() in ("none", "null", ""):
                return None
        else:
            try:
                return Config.convert_env_value(key, env_value, arg)
            except ValueError:
                continue
```

For `Union[str, None]`, `get_args` returns `(str, NoneType)` in declaration order, so the loop reaches `str` first. `convert_env_value(key, env_value, str)` returns `env_value` unconditionally (the `str` branch never raises `ValueError`), so the function returns before it ever tests the `NoneType` sentinel. The `none` / `null` / `""` to `None` coercion never runs for any `Optional[str]` field.

Affected fields (all `Union[str, None]` in `config/variables/base.py`): `AGENT_ROLE`, `REPORT_SOURCE`, `IMAGE_GENERATION_MODEL`.

It is reachable from a normal `.env`: `_set_attributes` (`config.py:71-75`) calls `convert_env_value` whenever `os.getenv(key) is not None`, which includes an empty string.

**To Reproduce**

Verified in a clean `python:3.12-slim` container with `pip install .` at HEAD `1a0a372829048090d153af99ccca2aca326be2ee`. Minimal pasteable repro:

```python
from typing import Union
from gpt_researcher.config.config import Config

for raw in ("none", "null", ""):
    print(repr(raw), "->", repr(Config.convert_env_value("AGENT_ROLE", raw, Union[str, None])))
```

Output:

```
'none' -> 'none'
'null' -> 'null'
'' -> ''
```

**Expected behavior**

Each call should return `None`, which is exactly what the `if env_value.lower() in ("none", "null", "")` branch is written to do.

Concrete consequence for `AGENT_ROLE`: with `AGENT_ROLE=none` set, `skills/writer.py:40/109/158` evaluates `self.researcher.cfg.agent_role or self.researcher.role`. Because `'none'` is a truthy string, the `or` does not fall back, and `'none'` is passed as `agent_role_prompt`, which becomes the system-message content in `actions/report_generation.py:41/92`. The model receives the literal system role `none` instead of the intended auto-generated role. (The empty-string case is masked downstream by the `or`, but it still leaves the field as `''` rather than the `None` default the code intends.)

**Environment**

- gpt-researcher HEAD `1a0a372` (`main`)
- Python 3.12, fresh `pip install .` in a clean container
- Not OS or browser specific (config/env parsing)

**Additional context**

A minimal fix is to check the `None` sentinel before iterating the non-`None` args, e.g. `if type(None) in args and env_value.lower() in ("none", "null", ""): return None`, then loop over the remaining args with the existing `try/except ValueError`. Happy to send a PR with a unit test if that is a welcome direction.

Disclosure: I found and verified this with AI-assisted tooling; the output above is what I actually ran.

Bartok9:
## Summary

- Coerce env strings none/null/empty to Python None for Optional config fields before str conversion can swallow them.
- AGENT_ROLE=none (and similar) now falls through writer role fallback instead of becoming the literal system role "none".

## Motivation

Closes #1899.

convert_env_value already intended to treat none/null/empty as None for Union[..., None] types, but get_args orders str before NoneType and the str branch returns unconditionally. The None-sentinel branch was therefore unreachable for every Optional[str] field (AGENT_ROLE, REPORT_SOURCE, IMAGE_GENERATION_MODEL).

## Verification

Real behavior after fix (same static method as issue repro):
- none/null/""/NONE/Null -> None
- researcher stays researcher
- Optional[int] none -> None; 42 -> 42

Unit tests: tests/test_convert_env_value_optional.py

Diff scope:
- gpt_researcher/config/config.py
- tests/test_convert_env_value_optional.py

### DIFF
--- a/gpt_researcher/config/config.py
+++ b/gpt_researcher/config/config.py
@@ -260,16 +260,19 @@ class Config:
         args = get_args(type_hint)
 
         if origin is Union:
-            # Handle Union types (e.g., Union[str, None])
+            # Handle Union types (e.g., Union[str, None] / Optional[str]).
+            # Check the None sentinel BEFORE non-None args: for Optional[str],
+            # str conversion never raises, so looping str-first permanently
+            # shadowed the none/null/"" → None branch (see issue #1899).
+            if type(None) in args and env_value.lower() in ("none", "null", ""):
+                return None
             for arg in args:
                 if arg is type(None):
-                    if env_value.lower() in ("none", "null", ""):
-                        return None
-                else:
-                    try:
-                        return Config.convert_env_value(key, env_value, arg)
-                    except ValueError:
-                        continue
+                    continue
+                try:
+                    return Config.convert_env_value(key, env_value, arg)
+                except ValueError:
+                    continue
             raise ValueError(f"Cannot convert {env_value} to any of {args}")
 
         if type_hint is bool:
--- a/tests/test_convert_env_value_optional.py
+++ b/tests/test_convert_env_value_optional.py
@@ -1,0 +1,25 @@
+"""Regression tests for Optional[str] env coercion (issue #1899)."""
+from typing import Optional, Union
+
+import pytest
+
+from gpt_researcher.config.config import Config
+
+
+@pytest.mark.parametrize(
+    "raw",
+    ["none", "null", "", "NONE", "Null"],
+)
+def test_optional_str_coerces_none_sentinels(raw):
+    assert Config.convert_env_value("AGENT_ROLE", raw, Union[str, None]) is None
+    assert Config.convert_env_value("AGENT_ROLE", raw, Optional[str]) is None
+
+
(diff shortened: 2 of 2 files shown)


assafelovic:
Superseded by #1910 ✅

PR #1903, opened 2026-07-14
fix(retriever): guard CRW against malformed search items

Bartok9:
## Summary

- Harden CRWRetriever.search so non-dict / missing-URL payload items are skipped instead of KeyErroring on obj["url"].
- Preserve markdown-over-description body selection and accept href as a URL alias.

## Motivation

Malformed entries in fastCRW data[] currently raise KeyError during list comprehension, which the outer except converts into an empty result set for the whole query. One bad item should not discard valid results.

## Verification

Direct module load of gpt_researcher/retrievers/crw/crw.py (avoiding package-level json_repair import) with mocked POST:

- malformed mix [None, str, missing-url, url+desc, href+markdown] -> two clean href/body dicts
- Existing shape tests remain valid

## Diff scope

- gpt_researcher/retrievers/crw/crw.py
- tests/test_crw_retriever.py (test_search_skips_malformed_items)

### DIFF
--- a/gpt_researcher/retrievers/crw/crw.py
+++ b/gpt_researcher/retrievers/crw/crw.py
@@ -110,14 +110,21 @@ class CRWRetriever:
             sources = results.get("data", [])
             if not sources:
                 raise Exception("No results found with fastCRW API search.")
-            # Return the results
-            search_response = [
-                {
-                    "href": obj["url"],
-                    "body": obj.get("markdown") or obj.get("description", ""),
-                }
-                for obj in sources
-            ]
+            # Return the results. Tolerate non-dict / missing-URL items so one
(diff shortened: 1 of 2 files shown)


assafelovic:
Thanks @Bartok9! A CRW guard from your #1866 is already merged into #1907. This one overlaps/conflicts on `crw.py` — if it adds handling beyond #1866, please rebase onto the latest `main`; otherwise it can be closed as covered. 🙏

Bartok9:
🤖 Covered — #1907's CRW guard (`sources = results.get("data") or []` + per-item `if obj.get("url")` skip) covers this. Closing as covered. Thanks! 🙏

PR #1904, opened 2026-07-14
fix(retriever): guard BoCha against malformed result payloads

Bartok9:
## Summary

- Harden BoChaSearch.search for missing data.webPages.value structure and per-item missing url/name/snippet.
- Return only valid {title,href,body} rows instead of failing the whole window.

## Motivation

BoCha normalization used result["url"] / result["name"] / result["snippet"] without guards. API envelope drift or partial objects raise KeyError and drop all results for the query. Matches the defensive pattern already used on Serper/Tavily-adjacent Bartok9 PRs.

## Verification

Isolated load of bocha.py with mocked POST:
- mixed None + missing-url + good item -> one cleaned result
- empty/malformed envelope -> []

## Diff scope

- gpt_researcher/retrievers/bocha/bocha.py
- tests/test_bocha_retriever.py

### DIFF
--- a/gpt_researcher/retrievers/bocha/bocha.py
+++ b/gpt_researcher/retrievers/bocha/bocha.py
@@ -40,19 +40,39 @@ class BoChaSearch():
             "count": max_results
         }
 
-        response = requests.post(url, headers=headers, json=data)
+        try:
+            response = requests.post(url, headers=headers, json=data, timeout=30)
+            response.raise_for_status()
+            json_response = response.json()
+        except Exception as e:
+            logging.getLogger(__name__).error(
+                "BoCha search request failed: %s. Returning empty results.", e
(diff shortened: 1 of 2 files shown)


assafelovic:
Thanks @Bartok9! A BoCha fix from your #1844 is already merged into #1907. This one overlaps/conflicts on `bocha.py` — if it adds handling beyond #1844, please rebase onto the latest `main`; otherwise it can be closed as covered. 🙏

Bartok9:
🤖 Covered — #1907's BoCha guard walks `data.webPages.value` defensively and `.get()`s each field, which covers the malformed-payload case. My only delta was a non-dict-item skip; not worth a rebase over the consolidated version. Closing as covered. Thanks! 🙏

PR #1905, opened 2026-07-14
fix(retriever): guard Tavily against malformed result fields

Bartok9:
## Summary

- Harden TavilySearch.search normalization: skip non-dict/missing-URL items; accept content/snippet/body for body text.
- Prevents a single bad results[] element from KeyErroring and wiping the whole query.

## Motivation

Current code does obj["url"] / obj["content"] in a list comprehension. Partial API payloads raise and the broad except returns []. Same defensive contract as other retriever guards.

## Verification

Isolated module load + mocked _search with mixed bad/good rows -> only valid href/body pairs.

## Diff scope

- gpt_researcher/retrievers/tavily/tavily_search.py
- tests/test_tavily_result_guard.py

### DIFF
--- a/gpt_researcher/retrievers/tavily/tavily_search.py
+++ b/gpt_researcher/retrievers/tavily/tavily_search.py
@@ -131,10 +131,17 @@ class TavilySearch:
             sources = results.get("results", [])
             if not sources:
                 raise Exception("No results found with Tavily API search.")
-            # Return the results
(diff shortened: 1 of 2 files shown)


Bartok9:
Closing as duplicate of already-open #1869 (same Tavily malformed-result guard). Prefer consolidating review on that PR.

PR #1914, opened 2026-07-15
fix(searchapi): tolerate missing keys on organic results

Bartok9:
## Summary

- SearchApi retriever no longer KeyErrors on missing/null organic_results or partial organic hits missing title/link/snippet.
- One malformed hit can no longer abort the whole page of search results (same hardening pattern as SerpApi/Serper).

## Motivation

Sibling retrievers (serpapi, serper) already normalize API payloads with .get() so a single incomplete organic result does not throw KeyError and discard every other hit via the broad except. SearchApiSearch still indexed with strict key access on organic_results / title / link / snippet.

An error payload without organic_results, JSON null list, or a hit missing snippet/title raised KeyError, which the surrounding except Exception caught — returning [] even when other hits on the same response were valid.

## Verification

- python -m pytest tests/test_searchapi_malformed_results.py -v — 4 passed, 0 failed
- Fail-without-fix: bare result["link"] / ["organic_results"] KeyError under the same payloads
- Did NOT change: auth header path, engine defaults, max_results semantics beyond validity of list length

## Real behavior proof

```
tests/test_searchapi_malformed_results.py::test_searchapi_skips_results_missing_keys PASSED
tests/test_searchapi_malformed_results.py::test_searchapi_no_organic_results_returns_empty PASSED
tests/test_searchapi_malformed_results.py::test_searchapi_null_organic_results_returns_empty PASSED
tests/test_searchapi_malformed_results.py::test_searchapi_skips_youtube_links PASSED
```

### DIFF
--- a/gpt_researcher/retrievers/searchapi/searchapi.py
+++ b/gpt_researcher/retrievers/searchapi/searchapi.py
@@ -61,22 +61,33 @@ class SearchApiSearch():
             response = requests.get(encoded_url, headers=headers, timeout=20)
             if response.status_code == 200:
                 search_results = response.json()
-                if search_results:
-                    results = search_results["organic_results"]
-                    results_processed = 0
-                    for result in results:
-                        # skip youtube results
-                        if "youtube.com" in result["link"]:
-                            continue
-                        if results_processed >= max_results:
-                            break
-                        search_result = {
-                            "title": result["title"],
-                            "href": result["link"],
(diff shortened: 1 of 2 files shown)


assafelovic:

Bartok9:
Thanks @assafelovic — glad the report helped and that #1852's approach was the one that landed in #2074. Appreciated.

PR #1915, opened 2026-07-15
fix(scraper): harden TavilyExtract against partial API payloads

Bartok9:
## Summary

- TavilyExtract no longer KeyErrors on missing `failed_results` / `results` / `raw_content`.
- `session=None` (constructor default) no longer AttributeErrors on `session.get` before returning valid extract content.

## Motivation

`scrape()` used strict dict indexing (`response['failed_results']`, `response['results'][0]['raw_content']`) and always called `self.session.get(...)`. Partial extract payloads or callers that only need text (no session) threw into the broad except and returned empty even when `raw_content` was present.

## Verification

- `python -m pytest tests/test_tavily_extract_malformed.py -v` — **6 passed**, 0 failed
- Did NOT change: Tavily client construction, failure when `failed_results` is non-empty, empty return contract on total failure

## Real behavior proof

```
tests/test_tavily_extract_malformed.py::test_successful_payload_returns_content_without_session PASSED
tests/test_tavily_extract_malformed.py::test_missing_failed_results_key_still_returns_content PASSED
tests/test_tavily_extract_malformed.py::test_null_results_returns_empty PASSED
tests/test_tavily_extract_malformed.py::test_empty_results_returns_empty PASSED
tests/test_tavily_extract_malformed.py::test_failed_results_present_returns_empty PASSED
tests/test_tavily_extract_malformed.py::test_missing_raw_content_key_returns_empty PASSED
```

### DIFF
--- a/gpt_researcher/scraper/tavily_extract/tavily_extract.py
+++ b/gpt_researcher/scraper/tavily_extract/tavily_extract.py
@@ -37,23 +37,37 @@ class TavilyExtract:
 
         try:
             response = self.tavily_client.extract(urls=self.link)
-            if response['failed_results']:
+            if not isinstance(response, dict):
                 return "", [], ""
 
-            # Parse the HTML content of the response to create a BeautifulSoup object for the utility functions
-            response_bs = self.session.get(self.link, timeout=4)
-            soup = BeautifulSoup(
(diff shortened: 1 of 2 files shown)


assafelovic:

Bartok9:

PR #1916, opened 2026-07-15
fix(scraper): guard process_scraped_data against partial items

Bartok9:
## Summary

- `process_scraped_data` uses safe dict access for status/content/url.
- Non-dict scrape entries are skipped; image_urls aggregation no longer assumes every item is a dict.

## Motivation

Post-scrape processing used strict key access. A single partial item without `status`/`content`/`url` raised KeyError mid-list and aborted processing of remaining successful scrapes.

## Verification

- `python -m pytest tests/test_process_scraped_data_guards.py -v` — **2 passed**
- Did NOT change: success path content extraction for well-formed items; non-success items still re-emitted as-is

## Real behavior proof

```
tests/test_process_scraped_data_guards.py::test_process_scraped_data_success_and_partial PASSED
tests/test_process_scraped_data_guards.py::test_process_scraped_data_missing_status_not_success PASSED
```

### DIFF
--- a/gpt_researcher/actions/web_scraping.py
+++ b/gpt_researcher/actions/web_scraping.py
@@ -35,8 +35,11 @@ async def scrape_urls(
         scraper = Scraper(urls, user_agent, cfg.scraper, worker_pool=worker_pool)
         scraped_data = await scraper.run()
         for item in scraped_data:
-            if 'image_urls' in item:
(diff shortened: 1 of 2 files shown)


assafelovic:

Bartok9:

PR #1917, opened 2026-07-15
fix(arxiv): guard null pdf_url/title and API exceptions

Bartok9:
## Summary

- ArxivSearch skips results with no usable href (pdf_url/entry_id) and defaults empty title/body.
- API/client exceptions return [] instead of crashing the research run.

## Motivation

Partial arxiv.Result metadata (None pdf_url/title/summary) produced unusable dicts or AttributeErrors. Uncaught exceptions from arxiv.Client.results() aborted the whole search path rather than degrading to empty results like sibling retrievers.

## Verification

- `python -m pytest tests/test_arxiv_null_fields.py -v` — **2 passed**
- Did NOT change: sort criterion validation; max_results wiring into arxiv.Search

## Real behavior proof

```
tests/test_arxiv_null_fields.py::test_skips_results_without_href_and_defaults_fields PASSED
tests/test_arxiv_null_fields.py::test_api_exception_returns_empty PASSED
```

### DIFF
--- a/gpt_researcher/retrievers/arxiv/arxiv.py
+++ b/gpt_researcher/retrievers/arxiv/arxiv.py
@@ -19,22 +19,32 @@ class ArxivSearch:
         :param max_results:
         :return:
         """
-
-        arxiv_gen = list(arxiv.Client().results(
-        self.arxiv.Search(
-            query= self.query, #+
-            max_results=max_results,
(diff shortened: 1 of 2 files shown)


assafelovic:

Bartok9:

PR #1918, opened 2026-07-15
fix(getxapi): skip non-dict tweets and non-list payloads

Bartok9:
## Summary

- GetXAPI requires `tweets`/`data` to be a list; non-list payloads return `[]`.
- Skips non-dict tweet entries and treats non-dict `author` as `{}`.

## Motivation

Sibling X retriever (Xquik) already uses `or {}` / null guards. GetXAPI iterated `tweets` and called `.get()` on each element; a string or null entry raised AttributeError into `search()`'s broad except, discarding every other valid tweet on the page.

## Verification

- `python -m pytest tests/test_getxapi_malformed.py -v` — **2 passed**
- Did NOT change: auth header, URL, engagement field mapping for well-formed tweets

## Real behavior proof

```
tests/test_getxapi_malformed.py::test_skips_non_dict_tweets_and_null_author PASSED
tests/test_getxapi_malformed.py::test_non_list_tweets_returns_empty PASSED
```

### DIFF
--- a/gpt_researcher/retrievers/getxapi/getxapi.py
+++ b/gpt_researcher/retrievers/getxapi/getxapi.py
@@ -61,10 +61,16 @@ class GetXAPISearch:
             data = json.loads(resp.read().decode("utf-8"))
 
         tweets = data.get("tweets") or data.get("data") or []
+        if not isinstance(tweets, list):
+            return []
         search_results = []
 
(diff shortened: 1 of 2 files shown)


assafelovic:

Bartok9:

PR #1919, opened 2026-07-15
fix(actions): move imports above _normalize_sub_queries to fix NameError on startup

CedricKhuaYanHan:
## What

Moves the import block in `gpt_researcher/actions/query_processing.py` to the top of the file.

## Why

bcc1eac0 added `_normalize_sub_queries` at the top of the file, above the imports.
Its `Any`/`List` annotations are evaluated at definition time, so importing the module raises:

```
```

This breaks server startup (`python -m uvicorn main:app`) on a fresh clone of `main`, i.e. the README quick-start path.

No behavior change - imports are just moved above the function that uses them.

## Verified

- `python -c "from backend.server.app import app"` imports cleanly
- Server starts and serves the frontend on `localhost:8000`

### DIFF
--- a/gpt_researcher/actions/query_processing.py
+++ b/gpt_researcher/actions/query_processing.py
@@ -1,6 +1,11 @@
 import json_repair
+import logging
+from typing import Any, List, Dict
 
 from gpt_researcher.llm_provider.generic.base import ReasoningEfforts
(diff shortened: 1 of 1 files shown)


assafelovic:

PR #1920, opened 2026-07-16
fix(searchapi): tolerate missing keys on organic results

Bartok9:
## Summary

- SearchApi no longer KeyErrors on missing/null organic_results or partial organic hits.
- Same hardening pattern as SerpApi/Serper.

## Motivation

Strict key access on organic_results/title/link/snippet discarded entire result sets when one hit was partial (caught by broad except → empty list).

## Verification

- .venv/bin/python -m pytest tests/test_searchapi_malformed_results.py -v — 4 passed

## Real behavior proof

4 tests PASSED (skip missing keys, no organic_results, null organic_results, youtube filter).


assafelovic:

Bartok9:
Thanks @assafelovic — glad the report helped and that #1852's approach was the one that landed in #2074. Appreciated.

PR #1921, opened 2026-07-16
fix(bocha): skip non-dict and href-less search hits

Bartok9:
## Summary

- BoChaSearch skips non-dict organic rows and results without a URL instead of AttributeError/empty-href entries.
- Non-list webPages.value returns [].

## Motivation

The walk for data.webPages.value was defensive, but the row loop assumed every element was a mapping and always emitted href even when blank. Partial API rows can break the shared retriever contract for consumer code that expects real URLs.

## Verification

- .venv/bin/python -m pytest tests/test_bocha_malformed_results.py -v — 2 passed

## Real behavior proof

tests/test_bocha_malformed_results.py both PASSED.


assafelovic:

Bartok9:

PR #1922, opened 2026-07-16
fix(brave): guard non-dict web/results payloads

Bartok9:
## Summary

- BraveSearch treats non-dict JSON, null web, or non-list results as empty rather than AttributeError.
- Skips non-dict organic rows.

## Verification

- .venv/bin/python -m pytest tests/test_brave_malformed_results.py -v — 3 passed

## Real behavior proof

All three malformed-payload tests PASSED.


assafelovic:

Bartok9:

PR #1923, opened 2026-07-16
fix(crw): skip non-dict sources in search results

Bartok9:
## Summary

- CRWRetriever skips non-dict source rows and requires data to be a list.
- Avoids AttributeError on obj.get when API mixes types into data.

## Verification

- .venv/bin/python -m pytest tests/test_crw_malformed_results.py -v — 2 passed

## Real behavior proof

Both crw malformed-result tests PASSED.


assafelovic:

Bartok9:

PR #1924, opened 2026-07-16
Add regional endpoints for provider integrations

octo-patch:
## Summary

- Add regional endpoint selection for the existing provider integration.
- Add a compatible protocol adapter with validated base URL handling.
- Record target model metadata, update configuration dependencies, and document both endpoint regions.

## Validation

- `PYTHONPATH=. python -m pytest tests/test_minimax_provider.py tests/test_nebius_provider.py tests/test_llm_max_tokens.py tests/test_llm_usage_tracking.py -q`
- `PYTHONPATH=. python -m compileall -q gpt_researcher/llm_provider/minimax.py gpt_researcher/llm_provider/generic/base.py gpt_researcher/memory/embeddings.py tests/test_minimax_provider.py`
- `git diff --check`


assafelovic:
Reviewed — this no longer applies to `main`, partly because of a change I just made: #2075 removed the dead `[tool.poetry.dependencies]` block that this PR also edits.

When I tested it against the pre-#2075 tree it applied and the suite passed, so the work itself looks sound — this is a mechanical rebase, not a rethink.

One note for the rebase: dependencies now live **only** in the PEP 621 `[project]` table. The `[tool.poetry]` block was dead config — the built wheel took every `Requires-Dist` from `[project]` — so any dependency change should go there.

octo-patch:
Thanks for the review. I updated the branch to current `main`, kept the dependency in the PEP 621 `[project]` table, and pushed a new merge commit. I ran the focused MiniMax provider tests on Python 3.11 (4 passed), plus metadata, syntax, and diff checks.

PR #1931, opened 2026-07-17
fix(bing): skip non-dict results and empty URLs

Bartok9:
## Summary

- BingSearch.search() assumed every webPages.value entry was a dict with a usable url.
- Null/string/incomplete rows are skipped; missing title/snippet default to empty string so one bad result cannot abort the research run.

## Motivation

Sibling retrievers already guard with isinstance(result, dict) before .get(). Bing did not, so a single non-dict value raised AttributeError mid-loop.

## Verification

```
python3 -m pytest tests/test_bing_retriever_malformed.py -v
# 3 passed
```

## Test plan

- [x] Unit tests cover missing keys, non-dict/null rows, empty webPages
- [ ] CI green

### DIFF
--- a/gpt_researcher/retrievers/bing/bing.py
+++ b/gpt_researcher/retrievers/bing/bing.py
@@ -79,17 +79,25 @@ class BingSearch():
             self.logger.warning(f"No search results found for query: {self.query}")
             return []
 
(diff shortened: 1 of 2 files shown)


assafelovic:

Bartok9:

PR #1932, opened 2026-07-17
fix(bocha): skip non-dict results and empty URLs

Bartok9:
## Summary

- BoChaSearch already resilient on network/JSON/wrapper shape, but still called .get on every webPages.value element.
- Skip non-dict/null rows and empty URLs; default missing name/snippet to empty strings.

## Motivation

Parity with Google/Serper/DuckDuckGo retrievers. One bad list element must not AttributeError the whole research loop.

## Verification

```
python3 -m pytest tests/test_bocha_error_handling.py -v
# 4 passed
```

## Test plan

- [x] Local unit tests
- [ ] CI green

### DIFF
--- a/gpt_researcher/retrievers/bocha/bocha.py
+++ b/gpt_researcher/retrievers/bocha/bocha.py
@@ -58,13 +58,22 @@ class BoChaSearch():
         ).get("value") or []
         search_results = []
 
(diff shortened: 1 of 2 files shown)


assafelovic:

Bartok9:

PR #1933, opened 2026-07-17
fix(serpapi): skip non-dict organic_results rows

Bartok9:
## Summary

- SerpApi already defaults missing organic_results to [] and skips empty links.
- Also skip non-dict rows, coerce organic_results when not a list, and default title/snippet with or "".

## Motivation

A single non-dict organic result raised AttributeError mid-search; matches Bing/BoCha/Google guards.

## Verification

```
python3 -m pytest tests/test_serpapi_malformed_results.py -v
# 2 passed
```

## Test plan

- [x] Local unit tests
- [ ] CI green


assafelovic:

Bartok9:

PR #1934, opened 2026-07-17
fix(custom): skip non-dict custom retriever items

Bartok9:
## Summary

- CustomRetriever already returns [] for HTTP/JSON/non-list failures.
- Also filter individual list elements: skip non-dict/null rows and items without url/href; normalize raw_content from body when needed.

## Motivation

Downstream code does result.get(...) on custom search hits. A list like [null, "x", {good}] used to pass validation as a list and crash later. Surface only dict hits with a usable URL.

## Verification

```
python3 -m pytest tests/test_custom_retriever_guard.py -v
# 5 passed
```

## Test plan

- [x] Local unit tests
- [ ] CI green

### DIFF
--- a/gpt_researcher/retrievers/custom/custom.py
+++ b/gpt_researcher/retrievers/custom/custom.py
@@ -68,4 +68,21 @@ class CustomRetriever:
                 f"{type(payload).__name__}"
             )
             return []
-        return payload
+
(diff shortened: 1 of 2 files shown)


assafelovic:

Bartok9:

PR #1935, opened 2026-07-17
fix(query_processing): add missing typing imports

FlorianFranzen:
This was needed on my end to make the multi agent ag2 backend run. Not sure why nobody else has this problem. This was using python 3.13.


assafelovic:

PR #1937, opened 2026-07-18
fix(xquik): skip non-list tweets and non-dict rows

Bartok9:
## Summary

- Xquik API payloads sometimes return a non-list `tweets` value or sparse non-dict rows.
- Those shapes used to AttributeError / iterate a string inside the worker loop; the broad `search()` except swallowed the failure and dropped **every** result.
- Guard: require `tweets` to be a list, skip non-dict rows, keep valid tweets.

## Motivation

Hardening sibling of the null-field work in `tests/test_xquik_null_fields.py`. Explicit JSON nulls were already handled; non-list envelopes and non-dict rows were not.

## Verification

Manual import of `xquik.py` (avoids package `json_repair` requirement):
- `tweets: null` → `[]`
- `tweets: "not-a-list"` → `[]`
- mixed rows with one valid tweet → 1 result with `@ok`

```
python3 - <<'PY'
# load module directly; assert non-list / non-dict guards
...
OK manual proof
PY
```

### DIFF
--- a/gpt_researcher/retrievers/xquik/xquik.py
+++ b/gpt_researcher/retrievers/xquik/xquik.py
@@ -73,10 +73,17 @@ class XquikSearch:
         # default and crash slicing / attribute access below — which the
         # broad except in search() would swallow, silently dropping every
         # result. Mirrors the sibling GetXAPI retriever.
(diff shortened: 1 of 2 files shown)


assafelovic:

Bartok9:

PR #1938, opened 2026-07-18
fix(openalex): tolerate non-dict OA/primary locations

Bartok9:
## Summary

- `_pick_href` assumed `best_oa_location` / `primary_location` were always dicts.
- Non-dict envelopes (list/string) raised `AttributeError` while normalizing a search hit and dropped the work.
- Require dict locations; otherwise fall through to the OpenAlex work `id`.

## Motivation

Defensive hardening for partial OpenAlex payloads. Truthy non-dict values defeat the existing `or {}` pattern (`[{...}] or {}` stays a list).

## Verification

Direct load of `openalex.py`:
```
_pick_href({..., best_oa_location: [{}], primary_location: "x"}) -> work id
```


assafelovic:

Bartok9:

PR #1939, opened 2026-07-18
fix(scraper): skip None/non-dict gather results in run

Bartok9:
## Summary

- `Scraper.run` filtered gather results with `content["raw_content"]`.
- A single `None` / non-dict worker payload crashed the whole batch filter, discarding successful scrapes.
- Require `isinstance(content, dict)` and use `.get("raw_content")`.

## Motivation

Parallel extractors can return unexpected shapes under partial failures. The batch should degrade gracefully instead of raising mid-filter.

## Verification

```
python3 tests/test_scraper_run_guards.py
# Ran 2 tests ... OK
```

### DIFF
--- a/gpt_researcher/scraper/scraper.py
+++ b/gpt_researcher/scraper/scraper.py
@@ -69,7 +69,14 @@ class Scraper:
             *(self.extract_data_from_url(url, self.session) for url in self.urls)
         )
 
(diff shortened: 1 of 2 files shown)


assafelovic:

Bartok9:

PR #1940, opened 2026-07-18
fix(pubmed): guard non-dict esearchresult and non-list idlist

Bartok9:
## Summary

- `_search_articles` chained `.get` on `esearchresult` assuming a dict.
- Null/list envelopes raised AttributeError; non-list `idlist` broke `len`/iteration.
- Restrict to dict + list shapes; otherwise return `[]`.

## Motivation

NCBI error JSON is not always the happy-path shape. Fail closed to an empty id list rather than crashing lookup.

## Verification

Direct module load:
- `esearchresult: null` → `[]`
- `idlist: "1,2,3"` → `[]`


assafelovic:

Bartok9:

PR #1944, opened 2026-07-18
fix: scraper ingests anti-bot challenge pages as content, word-list dumps dominate context

MrSampson:
## Summary

Two related content-quality problems found running this against real research queries in production, both returning HTTP 200 with a real (often large) body, so neither `Scraper.extract_data_from_url`'s `except Exception` path nor its existing `len(content) < 100` check catches either:

1. **Authoritative sources are correctly found by search, then return anti-bot/challenge pages when scraped** -- e.g. Anubis proof-of-work challenges (`inria.hal.science` and other Anubis-fronted sites), Cloudflare ("Attention Required!"), ResearchGate's "Temporarily Unavailable". Those challenge pages are then ingested as if they were the article's real content -- in one case, the report's citation for a specific paper was grounded entirely in Anubis's own challenge-page prose, not the paper.
2. **Raw word-list/vocabulary-dump `.txt` files scrape cleanly and then dominate the returned context** -- e.g. `cs.princeton.edu/.../words-333333.txt` (4.96M chars), `downloads.cs.stanford.edu/nlp/data/jiwei/data/vocab_wiki.txt` (1.68M chars). These lexically match almost any query, so they rank on everything; in one run, a single vocab dump contributed more chunks to the returned context than any legitimate source.

## Fix

After a scrape completes, `raw_content` is nulled out (the same shape the function already returns for too-short content) when it matches either check:

- **Block-page detection**: a small set of known anti-bot/challenge-page markers (Anubis, Cloudflare, ResearchGate, generic "verify you are human" phrasing), case-insensitive substring match.
- **Word-list detection**: large content (>200K chars) with near-zero sentence-ending punctuation density. Size alone isn't used as the signal -- legitimate long documents exist -- only the near-total absence of sentence structure combined with real size, which is what actually distinguishes a word-list dump from any real prose (even dense technical writing has a period every ~100-200 characters; a plain word list has essentially none).

Downstream code already drops `raw_content=None` entries (`Scraper.run`'s filter, and `_search_relevant_source_urls`'s classification), so both cases are then treated exactly like "this URL failed to scrape," and gpt-researcher tries another source instead of treating either as real content.

## Test plan

`tests/test_scraper_content_quality.py`:
- Unit tests for the two detection helpers directly, including false-positive checks against long legitimate prose (>200K chars, but full of periods) and short normal content -- both must survive untouched.
- An integration-style test driving the full `extract_data_from_url` path with a stubbed scraper backend, confirming the method-level contract (not just the helper functions in isolation).
- Ran alongside the existing scraper test suite (`test_scraper_extract_title.py`, `test_scraper_get_scraper.py`, `test_arxiv_scraper.py`): 13 passed, no regressions.

(Note: verifying this locally required #1943 -- the currently-published release doesn't import at all -- applied on top; this PR's diff itself doesn't touch that file.)

### DIFF
--- a/gpt_researcher/scraper/scraper.py
+++ b/gpt_researcher/scraper/scraper.py
@@ -31,19 +31,27 @@ from . import (
 # These pages return HTTP 200 with real (often large) HTML bodies, so
 # neither an exception nor the short-content check below catches them --
 # without this, a block/challenge page is ingested as if it were the
-# article's real content. Not exhaustive; add more as new blockers are
-# observed in practice.
+# article's real content. Each marker is anchored specifically enough to
+# avoid colliding with ordinary prose (e.g. "researchgate - temporarily
+# unavailable", not the bare phrase "temporarily unavailable", which a
+# legitimate article about an unrelated outage could plausibly contain).
+# Not exhaustive; add more as new blockers are observed in practice.
 _BLOCK_PAGE_MARKERS = (
     "anubis uses a proof-of-work scheme",  # HAL and other Anubis-fronted sites
     "making sure you're not a bot",
     "checking your browser before accessing",
     "enable javascript and cookies to continue",
-    "temporarily unavailable",  # ResearchGate's specific wording
+    "researchgate - temporarily unavailable",  # ResearchGate's specific wording
     "please verify you are a human",
     "attention required! | cloudflare",
     "sorry, you have been blocked",
-    "just a moment...",
 )
+
+# Block pages are the entire response body -- a "please wait" message, not
+# a real article -- so they're always short and always appear at the very
+# start of the content. Checking only a prefix avoids lower-casing and
+# scanning multi-megabyte legitimate documents on every scrape.
+_BLOCK_PAGE_CHECK_PREFIX_LEN = 5_000
 
 # Word-list/vocab dumps (plain lists of unrelated words, no prose) scrape
 # cleanly and can dominate a report's context, since they lexically match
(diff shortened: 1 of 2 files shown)


MrSampson:
Pushed a follow-up commit addressing review feedback:

- Anchored "temporarily unavailable" to the full ResearchGate wording and dropped the standalone "just a moment..." marker (too generic) — both were reproduced false-positiving on an ordinary news article. Cloudflare detection stays covered by the three other, more specific markers.
- The word-list check now also counts fullwidth CJK sentence terminators (。！？), not just ASCII `. ! ?` — otherwise long-form Chinese/Japanese/Korean prose was misclassified as a word-list dump purely for lacking Latin punctuation. Reproduced with a 440K-char Chinese sample.
- Block-page markers are now checked against only the first 5,000 chars (block pages are always short and always at the very start), which also avoids scanning multi-megabyte legitimate documents on every scrape.
- Deduplicated the four reject-shape return blocks into one `_reject()` helper and removed a pre-existing unreachable dead-code check while touching this area.

New regression tests cover both reproduced false positives plus prefix-window boundary behavior. All 15 tests in `tests/test_scraper_content_quality.py` pass, plus the existing scraper suite (13 tests) with no regressions.

assafelovic:

FlorianFranzen:
@MrSampson Please be aware that despite what the LLM generated comment wants you to believe, your change has not been attributed with the appropriate `Co-Authored-By:` byline when it was merged upstream.

PR #1946, opened 2026-07-19
fix(serper): guard non-dict JSON search payloads

Bartok9:
## Summary
- Treat non-dict Serper JSON bodies (list/str/number) as empty results instead of calling `.get("organic")`.
- Keep existing empty-list behavior for null/unparseable bodies and skip non-dict organic rows.

## Motivation
`search()` always builds results via `search_results.get("organic")`. A 200 response that still `json.loads` to a list/string raises `AttributeError`, which aborts the whole research run instead of degrading to no Serper hits.

## Verification
Without fix (simulated): AttributeError 'list' object has no attribute 'get'
With fix: `python3 -m pytest tests/test_serper_non_dict_payload.py -q` → 3 passed

## Real behavior proof
- Payload `[1]` → `[]` (no crash)
- Normal `{"organic":[...]}` still returns normalized href/title/body rows

## Test plan
- [x] pytest tests/test_serper_non_dict_payload.py
- [ ] CI green on this PR

### DIFF
--- a/gpt_researcher/retrievers/serper/serper.py
+++ b/gpt_researcher/retrievers/serper/serper.py
@@ -112,7 +112,7 @@ class SerperSearch():
             search_results = json.loads(resp.text)
         except Exception:
             return []
-        if search_results is None:
+        if search_results is None or not isinstance(search_results, dict):
             return []
 
         results = search_results.get("organic") or []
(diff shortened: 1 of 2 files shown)


assafelovic:

PR #1947, opened 2026-07-19
fix(tavily): guard non-dict API responses in search()

Bartok9:
## Summary
- Require Tavily `_search` payloads to be a dict with a list `results` before iterating.
- Non-dict JSON or a non-list `results` degrades to `[]` via the existing exception handler.

## Motivation
After a successful HTTP call, `search()` did `results.get("results", [])`. List/scalar JSON bodies raise `AttributeError` and treat the failure path inconsistently with other retrievers.

## Verification
`python3 -m pytest tests/test_tavily_non_dict_response.py -q` → 3 passed

## Test plan
- [x] unit tests for list payload, non-list results, happy path
- [ ] CI green


assafelovic:

PR #1948, opened 2026-07-19
fix(firecrawl): allow scrape when session is None

Bartok9:
## Summary
- Skip the secondary HTML/image pass when `session` is `None` instead of calling `self.session.get`.
- Swallow image-pass errors so markdown content still returns successfully.

## Motivation
`FireCrawl(link, session=None).scrape()` crashed with `AttributeError: 'NoneType' object has no attribute 'get'` after a successful Firecrawl SDK scrape whenever callers omitted the shared `requests.Session`.

## Verification
`python3 -m pytest tests/test_firecrawl_null_session.py -q` → 1 passed

## Test plan
- [x] unit test session=None happy path
- [ ] CI green


assafelovic:

PR #1949, opened 2026-07-19
fix(scraper): tolerate invalid Content-Length in BS fetch

Bartok9:
## Summary
- Parse `Content-Length` defensively; invalid values no longer raise `ValueError` during `_fetch`.
- Oversized pages (valid integer length) are still skipped.

## Motivation
Some CDNs/proxies return non-numeric or blank `Content-Length` values. `int(content_length)` aborted the fetch path and bubbled into scraper error handling instead of parsing a still-valid body.

## Verification
`python3 -m pytest tests/test_bs_content_length_invalid.py -q` → expected pass

## Test plan
- [x] unit test invalid header
- [ ] CI green


assafelovic:

PR #1951, opened 2026-07-19
fix: detect and recover extensionless PDFs routed to the wrong scraper

MrSampson:
## Summary

`Scraper.get_scraper()` picks a backend from the URL's suffix, but PDFs served from institutional repository download endpoints (DSpace/EPrints-style, e.g. `scholarspace.manoa.hawaii.edu/bitstreams/<uuid>/download`) have no `.pdf` suffix. Those fall through to a text/HTML scraper, which has no way to decode the PDF body and returns its raw bytes (FlateDecode streams, xref tables, `/Annot` objects) as if it were the page's real text — silently: no exception, a normal-looking `content_length`, just unusable content passed on to synthesis.

This matters specifically because institutional repositories are where a lot of the highest-value open-access primaries live for research/regulatory queries — theses, preprints, university-hosted studies — so this bug hits exactly the sources that matter most, while working correctly for publisher PDFs that do end in `.pdf`.

## Fix

After a scrape completes, check whether the content looks like unextracted PDF binary — starts with `%PDF-`, or contains at least two independent PDF structural tokens (`endobj`/`endstream`/`/FlateDecode`/`xref`/`trailer` — syntax that essentially never occurs, even coincidentally, in real prose). If so, and the originally-dispatched backend wasn't already `PyMuPDFScraper`, retry the same URL directly with `PyMuPDFScraper`. Recovering real content this way is a better outcome than discarding the source outright.

## Verification against the real, reported-broken URL (not just synthetic test data)

Fetched `https://scholarspace.manoa.hawaii.edu/bitstreams/5b99bd69-e789-4864-9cd8-4fe3d912b55a/download` and ran it through the same `BeautifulSoup` → `clean_soup` → `get_text_from_soup` path `Scraper` actually uses:

- Reproduces the exact reported symptom: **1,346,436** "extracted" characters of raw PDF binary — matches the `content_length` in the bug report byte-for-byte.
- Detection correctly fires on this real content.
- Running the same file through `PyMuPDFLoader` (what the retry path uses) recovers **61,260** characters of clean, readable text — the actual paper ("Detection of Contradictions and Inconsistencies in Regulatory Documents using Prompt-Engineering").

## Test plan

`tests/test_scraper_pdf_retry.py`:
- Unit tests for the detection helper, including a false-positive guard (a single incidental word like "trailer" appearing in ordinary prose must not alone trigger it — two independent structural tokens are required).
- Integration-style tests driving the full `extract_data_from_url` retry path with both the originally-dispatched backend and `PyMuPDFScraper` replaced by stubs: successful recovery, retry-also-fails (falls back to a normal fetch-failure shape rather than passing binary through), and the already-correct-backend case (no retry loop).
- Ran alongside the existing scraper suite (`test_scraper_extract_title.py`, `test_scraper_get_scraper.py`, `test_arxiv_scraper.py`): 20 passed, no regressions.

### DIFF
--- a/gpt_researcher/scraper/scraper.py
+++ b/gpt_researcher/scraper/scraper.py
@@ -27,6 +27,29 @@ from . import (
     WebBaseLoaderScraper,
 )
 
+# get_scraper() picks a backend from the URL's suffix, but PDFs served from
+# institutional repositories (DSpace/EPrints-style download endpoints) are
+# routed by content type, not a file extension -- e.g.
+# "scholarspace.manoa.hawaii.edu/bitstreams/<uuid>/download" has none. Those
+# fall through to a text/HTML scraper, which has no way to decode the PDF
+# body and returns its raw bytes (FlateDecode streams, xref tables, /Annot
+# objects) as if it were the page's real text -- silently: no exception, a
+# normal-looking content_length, just unusable content.
+#
+# These tokens are PDF's own internal structural syntax; they essentially
+# never occur, even coincidentally, in real prose. Two independent hits is
+# enough to be confident this is raw PDF, not text that happens to contain
+# one of these words in isolation.
+_PDF_STRUCTURE_MARKERS = ("endobj", "endstream", "/FlateDecode", "xref", "trailer")
+_MIN_PDF_MARKER_HITS = 2
+
+
+def _looks_like_unextracted_pdf(text: str) -> bool:
+    if text.startswith("%PDF-"):
+        return True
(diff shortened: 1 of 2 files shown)


assafelovic:
Landed in `main` via #2074 — commit `4fb2ccbd` (*fix: recover extensionless PDFs routed to the wrong scraper*).

**Note:** this conflicted with #1944 in `scraper.py`. The two turned out to be complementary rather than competing — block-page/word-list detection vs unextracted-PDF retry — so both were kept and the checks run in sequence.

PR #1952, opened 2026-07-19
fix: cap SearxSearch/Duckduckgo snippet length so real pages actually get scraped

MrSampson:
## Summary

`gpt_researcher.skills.researcher._search_relevant_source_urls()` treats any search result whose `raw_content`/`body` exceeds 100 characters as already-fetched full text — a heuristic meant for retrievers (e.g. PubMed Central) that genuinely return full article text inline. `SearxSearch` and `Duckduckgo` both populate `"body"` with an ordinary search-result snippet, which routinely exceeds 100 characters, so every result from either retriever was wrongly treated as already-fetched and the real page was never actually scraped — `get_source_urls()` always returned `[]`, and report citations weren't grounded in anything actually fetched (in one observed case, a citation named the wrong CELEX number for the regulation it referenced).

## Fix

Cap `body` at 100 characters at the retriever source (`SearxSearch.search`/`Duckduckgo.search`), rather than patching `_search_relevant_source_urls()`'s classification logic itself — this way any future upstream change to that function (bug fixes, new bookkeeping) is inherited automatically rather than shadowed by a stale reimplementation. The truncated snippet is only ever used for the classification decision and early-stage sub-query planning; once a URL takes the real scrape path, gpt-researcher fetches and uses the actual page content, not this snippet.

This has been running in production (as an external monkeypatch, prior to this PR) for about two days at time of writing, with confirmed improvement: retrieval quality — including correctly finding and citing primary sources — improved markedly once this was in place.

## Test plan

- `tests/test_searx_malformed_results.py` / `tests/test_duckduckgo_normalize.py`: added a case each asserting a 250-char snippet is capped to exactly 100 chars, alongside the existing malformed-result-handling tests for both retrievers (both still pass unchanged, since their existing test snippets are all well under 100 chars).

### DIFF
--- a/gpt_researcher/retrievers/duckduckgo/duckduckgo.py
+++ b/gpt_researcher/retrievers/duckduckgo/duckduckgo.py
@@ -1,6 +1,23 @@
 from itertools import islice
 from ..utils import check_pkg
 
+# gpt_researcher.skills.researcher._search_relevant_source_urls() treats
+# any search result whose raw_content/body exceeds 100 characters as
+# already-fetched full text -- a heuristic meant for retrievers (e.g.
+# PubMed Central) that genuinely return full article text inline. ddgs
+# populates "body" with an ordinary search-result snippet, which routinely
+# exceeds 100 characters, so without this cap every result is wrongly
+# treated as already-fetched and the real page is never actually scraped
+# (get_source_urls() then returns [] and reports ship with zero verifiable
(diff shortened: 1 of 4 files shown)


assafelovic:

FlorianFranzen:
@MrSampson Please be aware that despite what the LLM generated comment wants you to believe, your change has not been attributed with the appropriate `Co-Authored-By:` byline when it was merged upstream.

PR #1954, opened 2026-07-19
fix(curator): recover fenced LLM JSON instead of dropping curation

### LINKED ISSUE #1953: Source curation silently degrades on markdown-fenced LLM output (curator uses json.loads, not the repo's json_repair)
### Summary

`SourceCurator.curate_sources` parses the model's response with plain `json.loads`:

```python
# gpt_researcher/skills/curator.py
curated_sources = json.loads(response)
```

LLMs routinely wrap the array in ` ```json ` fences or add a prose preamble despite the prompt's explicit "no markdown" instruction. `json.loads` raises on that, so the call drops into the `except` branch and returns the **full, uncurated** source set. The only signal is a `🚫 Source verification failed` log line, which misattributes the cause to the sources/model rather than to the parser.

### Impact

- **Severity: low-to-moderate.** No crash, no data loss, no security exposure — curation just silently no-ops and the report is built from all sources instead of the curated top-N.
- **Who's affected:** only users who opt in via `CURATE_SOURCES=True` (it defaults to `False`), *and* whose model doesn't perfectly obey the "no markdown" instruction.
- **Consequences when it fires:** (a) larger downstream context → higher token cost and latency in report generation; (b) quality dilution since noise isn't filtered; (c) the enabled feature does nothing.
- **Worst case:** on a less instruction-compliant model (e.g. many local / Ollama / smaller models), the output is fenced on *every* call, so curation is effectively **100% broken and silent** for that user — and the log blames "source verification," not the parser, so it's easy to miss.

### Reproduction

```python
import json
# Typical model output despite the prompt saying "no ```json":
json.loads('```json\n[{"Title":"A","Source":"https://a.example"}]\n```')
# -> json.decoder.JSONDecodeError

# Same for a prose prefix or a trailing comma.
```

With `CURATE_SOURCES=True` and any model that fences its JSON, `curate_sources` returns the uncurated input on every run.

### Why this looks like an oversight, not a deliberate choice

The rest of the codebase already parses LLM JSON robustly with the shipped `json-repair` dependency, precisely because models emit fenced/malformed output:

- `actions/query_processing.py` → `json_repair.loads(response)`
- `skills/deep_research.py` → `json_repair.loads(...)`
- `multi_agents/agents/utils/llms.py` → `parse_json_markdown(response, parser=json_repair.loads)`
- `actions/agent_creator.py` — explicitly falls back to `json_repair` "to recover…from malformed JSON responses"

`curator.py` is the lone LLM-JSON call site that uses bare `json.loads` with no recovery. `json-repair` is already a hard dependency (`requirements.txt`, `pyproject.toml`).

### Expected vs. actual

- **Expected:** a well-formed source list wrapped in fences is parsed and curation runs.
- **Actual:** it's discarded and curation silently falls back to the uncurated set.

### Proposed fix

Parse the response the same way `multi_agents` already does — no new dependency:

```python
from langchain_core.utils.json import parse_json_markdown
import json_repair
...
curated_sources = parse_json_markdown(response, parser=json_repair.loads)
```

The existing fallback on a genuinely unparseable response is intentionally left unchanged. Happy to open a PR with this plus a regression test.

### Environment

- Observed on `main` (current `gpt_researcher/skills/curator.py`).
- Model-agnostic; more frequent with models that don't strictly follow "return only JSON."

teionarr:
Fixes #1953.

### Problem

`SourceCurator.curate_sources` parses the model response with plain `json.loads`:

```python
curated_sources = json.loads(response)
```

Models routinely wrap the array in ` ```json ` fences or add a prose preamble despite the prompt's "no markdown" instruction. `json.loads` raises on that, so the call drops into the `except` branch and returns the **full, uncurated** source set — with only a `🚫 Source verification failed` log line that misattributes the failure to the sources/model rather than the parser. With `CURATE_SOURCES=True`, the feature silently no-ops on ordinary model output (and can be broken on *every* call for less instruction-compliant/local models).

### Why this is a gap, not intentional

The rest of the codebase already parses LLM JSON robustly with the shipped `json-repair` dependency:

- `actions/query_processing.py` → `json_repair.loads(response)`
- `skills/deep_research.py` → `json_repair.loads(...)`
- `multi_agents/agents/utils/llms.py` → `parse_json_markdown(response, parser=json_repair.loads)`
- `actions/agent_creator.py` explicitly falls back to `json_repair` "to recover…from malformed JSON responses"

The curator is the lone LLM-JSON call site using bare `json.loads` with no recovery.

### Fix

Parse the response the same way `multi_agents` already does. No new dependency (both `json_repair` and `parse_json_markdown` are already used in-repo).

The existing "fall back to uncurated sources" safety net is preserved with an explicit list-type guard: unlike `json.loads`, `json_repair` never raises — on unusable output it returns `""` or a `dict`. Without the guard, such a result would slip past the `except` branch and empty the report context. Rejecting a non-list result keeps the graceful degradation intact.

### Tests

Adds `tests/test_source_curator_json_parsing.py` covering three cases: a markdown-fenced response is now parsed to the ranked source (not the uncurated set); clean output still parses; and a non-list response falls back to the uncurated sources.

### Risk

Minimal — one parse call swapped, strictly more permissive than before, so no previously-passing input regresses.

### DIFF
--- a/gpt_researcher/skills/curator.py
+++ b/gpt_researcher/skills/curator.py
@@ -4,8 +4,10 @@ This module provides the SourceCurator class that evaluates and ranks
 research sources based on relevance, credibility, and reliability.
 """
 
-import json
 from typing import Dict, List, Optional
+
+import json_repair
+from langchain_core.utils.json import parse_json_markdown
 
 from ..actions import stream_output
 from ..config.config import Config
@@ -71,7 +73,20 @@ class SourceCurator:
                 cost_callback=self.researcher.add_costs,
             )
 
-            curated_sources = json.loads(response)
+            # LLMs frequently wrap the JSON in ```json fences or add prose
+            # despite the prompt's instructions, which plain json.loads cannot
+            # parse. Recover the payload the same way the rest of the codebase
+            # does (see actions/query_processing.py, multi_agents/agents/utils/
+            # llms.py) so a well-formed-but-fenced response is not discarded.
+            curated_sources = parse_json_markdown(response, parser=json_repair.loads)
+            # json_repair never raises: on unusable output it returns "" or a
+            # dict rather than a list. Guard so such a result still falls back
+            # to the uncurated sources below instead of emptying the context.
+            if not isinstance(curated_sources, list):
+                raise ValueError(
+                    f"expected a JSON list of sources, got "
+                    f"{type(curated_sources).__name__}"
+                )
             print(f"\n\nFinal Curated sources {len(source_data)} sources: {curated_sources}")
 
             if self.researcher.verbose:
--- a/tests/test_source_curator_json_parsing.py
+++ b/tests/test_source_curator_json_parsing.py
@@ -1,0 +1,90 @@
+"""Regression: SourceCurator must recover fenced/malformed LLM JSON.
+
+Models routinely wrap their JSON in ```json fences despite the curator
+prompt's "no markdown" instruction. Plain ``json.loads`` raises on that
+output, which sent ``curate_sources`` down its except-branch and silently
+returned the *full uncurated* source set — defeating the curation the user
+opted into via ``cfg.curate_sources``. The parser now matches the rest of
+the codebase (``parse_json_markdown`` + ``json_repair``), so a
+well-formed-but-fenced response is parsed instead of discarded.
+"""
+
+from types import SimpleNamespace
+from unittest.mock import AsyncMock
+
+import pytest
+
+from gpt_researcher.skills.curator import SourceCurator
+
+
+def _fake_researcher():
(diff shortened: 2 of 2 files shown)


assafelovic:

FlorianFranzen:
@teionarr Please be aware that despite what the LLM generated comment wants you to believe, your change has not been attributed with the appropriate `Co-Authored-By:` byline when it was merged upstream.

teionarr:
Thanks for flagging that @FlorianFranzen, appreciate the sharp eyes. Seems the one co-author who did make the trailer wasn't in a position to object. 🙂 It happens. the fix landed and the issue is closed, which is what I was after. @assafelovic added me into an update, not perfect for tracking contribution, but good in general.

PR #1955, opened 2026-07-20
fix(image_generator): recover fenced LLM JSON via json_repair

Bartok9:
## Summary

- `_plan_image_concepts` and `_parse_analysis_response` used bare `json.loads` (plus a brittle fence strip only when the response *starts* with fences).
- LLM replies that include a prose preamble or fenced JSON still failed parse and returned empty concepts/suggestions, so image generation silently no-ops.
- Switch both paths to `json_repair.loads` (already a hard dep used by `deep_research` / `query_processing`) and guard non-list/non-dict payloads.

## Motivation

Same class of failure as curator issue #1953 and the deep_research parsers: models frequently fence JSON despite "JSON only" prompts. Image generation already failed closed to `[]` on decode errors, so the feature looked configured but never drew. Aligning with `json_repair` restores the intended path without new dependencies.

## Verification

```text
$ .venv/bin/python -m pytest tests/test_image_generator_json_repair.py -q
3 passed
```

Bare `json.loads` on fenced input raises `JSONDecodeError`; `json_repair.loads` returns the array/object (covered by tests).

## Real behavior proof

```text
>>> import json
>>> json.loads('Here you go:\n```json\n[{"title":"Arch","prompt":"diagram"}]\n```')
JSONDecodeError
>>> import json_repair
>>> json_repair.loads('Here you go:\n```json\n[{"title":"Arch","prompt":"diagram"}]\n```')
[{'title': 'Arch', 'prompt': 'diagram'}]
```

### DIFF
--- a/gpt_researcher/skills/image_generator.py
+++ b/gpt_researcher/skills/image_generator.py
@@ -8,6 +8,8 @@ import asyncio
 import json
 import logging
 import re
+
+import json_repair
 from typing import Any, Dict, List, Optional, Tuple
 
 from ..actions.utils import stream_output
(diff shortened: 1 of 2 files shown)


assafelovic:

PR #1956, opened 2026-07-20
fix(mcp): recover fenced tool-selection JSON via json_repair

Bartok9:
## Summary

- `MCPToolSelector.select_relevant_tools` parsed the LLM reply with bare `json.loads` + a brittle `{...}` regex extract.
- Fenced / preambled JSON failed parse and dropped into `_fallback_tool_selection`, ignoring the model’s ranking.
- Use `json_repair.loads` (existing hard dep) and skip non-dict `selected_tools` rows.

## Motivation

Same silent-degrade class as image planning and deep_research JSON paths. When MCP tool selection is enabled, a fence should not erase the LLM ranking.

## Verification

```text
$ .venv/bin/python -m pytest tests/test_mcp_tool_selector_json_repair.py -q
2 passed
```

## Real behavior proof

```text
>>> import json
>>> json.loads('Pick:\n```json\n{"selected_tools":[{"index":1}]}\n```')
JSONDecodeError
>>> import json_repair
>>> json_repair.loads('Pick:\n```json\n{"selected_tools":[{"index":1}]}\n```')
{'selected_tools': [{'index': 1}]}
```

### DIFF
--- a/gpt_researcher/mcp/tool_selector.py
+++ b/gpt_researcher/mcp/tool_selector.py
@@ -6,6 +6,8 @@ Handles intelligent tool selection using LLM analysis.
 import asyncio
 import json
 import logging
+
+import json_repair
 from typing import List, Dict, Any, Optional
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


assafelovic:

PR #1957, opened 2026-07-20
fix(config): coerce list/dict env JSON via json_repair

Bartok9:
## Summary

- `Config.convert_env_value` used bare `json.loads` for `list` / `dict` env values.
- Hand-edited env (trailing commas, minor JSON glitches) raised at import/startup.
- Parse with `json_repair.loads` and reject wrong container types with `ValueError`.

## Motivation

Operators often put `RETRIEVER` domain lists and MCP/JSON knobs in `.env` by hand. A trailing comma should not take down the process when `json-repair` is already a core dependency.

## Verification

```text
$ .venv/bin/python -m pytest tests/test_convert_env_value_json_repair.py tests/test_convert_env_value_optional.py -q
6 passed
```

## Real behavior proof

```text
>>> import json
>>> json.loads('["a.com", "b.com",]')
JSONDecodeError
>>> import json_repair
>>> json_repair.loads('["a.com", "b.com",]')
['a.com', 'b.com']
```

### DIFF
--- a/gpt_researcher/config/config.py
+++ b/gpt_researcher/config/config.py
@@ -6,6 +6,7 @@ retrievers, and various operational parameters.
 """
 
 import json
+import json_repair
 import os
 import warnings
 from typing import Any, Dict, List, Type, Union, get_args, get_origin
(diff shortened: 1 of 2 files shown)


assafelovic:

PR #1958, opened 2026-07-20
fix(exa): return empty results on search API failures

Bartok9:
## Summary

- `ExaSearch.search` let `client.search` exceptions bubble and assumed `results.results` was iterable.
- One Exa failure could abort the whole retriever loop path depending on callers.
- Catch exceptions → `[]`; tolerate missing/non-list `results`.

## Verification

```text
$ .venv/bin/python -m pytest tests/test_exa_search_errors.py -q
3 passed
```


assafelovic:

PR #1961, opened 2026-07-20
fix: license per-claim grounding refusal in the default report prompt

MrSampson:
The default report prompt has no instruction for what to do when a cited source doesn't support a specific claim -- see linked issue for a worked example where an ad hoc custom prompt produced correct refusal behavior the default prompt never licensed.


assafelovic:
Reviewed — the actual change here is small and worth having, but it's buried.

GitHub reports **27,040 additions across 311 files**. Nearly all of that is a branch that picked up an unrelated working tree: `.claude/`, `.codex-plugin/`, `.mcp.json`, `.vscode/`, workflow files, and a `.claude/worktrees/crazy-curie-e43899` gitlink — that last one was itself a bug on `main` (it broke recursive clones and git-URL Docker builds) and was removed in #2074.

The fix described in the title looks like roughly twenty lines. Rebasing onto `main` and keeping only that would make this reviewable and, most likely, mergeable.

Worth saying plainly: **the content concerns are real and I'd like these to land.** Your #1944, #1951 and #1952 all merged in #2074 and were among the highest-value fixes in the batch — the scraper quality work in particular. This is a packaging problem, not a judgement on the work.

**Also: this targets `master`, which is no longer the trunk.** `main` is the default branch and is now 160 commits ahead; `master` last moved in April. Retargeting to `main` is needed regardless of anything else.

CI now runs on every PR — `imports` and `unit` across Python 3.11–3.14, plus `collect` — so once it's rebased you'll get a direct answer.

PR #1962, opened 2026-07-21
fix(searchapi): guard non-dict organic_results rows

Bartok9:
## Summary

Harden SearchApiSearch.search so partial or malformed SearchApi JSON no longer crashes the result walk with KeyError TypeError.

## Motivation

organic_results was accessed with bare indexing and each row used required keys for link title snippet. Error stubs list envelopes non-dict rows and rows missing link aborted the entire search path.

## Changes

- Require JSON object payload and list organic_results
- Skip non-dict rows and empty links
- Use get for title and snippet

## Real behavior proof

python3 -m pytest tests/test_searchapi_organic_guards.py -q
4 passed

## Test plan

- Unit tests cover non-dict rows missing organic key non-list organic non-dict JSON top-level
- Author Bartok9


assafelovic:

Issue #1945, opened 2026-07-18
No CI workflow actually runs the test suite (pytest never invoked)


MrSampson:
## Summary

None of this repo's 5 GitHub Actions workflows (`build.yml`, `deploy.yml`, `docker-build.yml`, `llm-costs.yml`, `plugin-quality-gate.yml`) invoke `pytest` — confirmed by grepping all five for the string. There's a real, non-trivial `tests/` directory, but nothing in CI actually runs it on pushes or PRs.

## Why this matters (concrete example, not hypothetical)

Commit `bcc1eac0` (2026-06-27) introduced `_normalize_sub_queries` in `gpt_researcher/actions/query_processing.py` with a typing import placed *after* the function that uses it — an import-ordering bug that makes the entire package fail with `NameError: name 'Any' is not defined` on module load, under any Python version without PEP 649/749 lazy annotation evaluation (i.e. everything before 3.14). This shipped all the way to the currently-published `0.16.0` release — the package is completely unimportable as published right now (see #1943, opened separately as a standalone hotfix given the severity).

The same commit that introduced the bug also added `tests/test_sub_query_normalization.py`, which does `from gpt_researcher.actions.query_processing import _normalize_sub_queries` at module scope — i.e. a regression test for exactly this import path already exists and would have caught this immediately. It didn't catch it because nothing runs it.

## Suggested fix

Wire `pytest` into at least one existing workflow (or add a new one) that runs on pull requests and pushes to `main`. Given the size of the dependency tree, even a fast subset (the existing lightweight `unittest`-based tests like `tests/test_scraper_get_scraper.py`, `tests/test_scraper_extract_title.py`, `tests/test_arxiv_scraper.py`, `tests/test_sub_query_normalization.py` — none of which need network access or API keys) would have caught this specific regression, if a full-suite run isn't currently feasible for cost/secrets reasons.

Happy to help with a PR for this if useful direction/constraints are given (e.g. which secrets are safe to expose to PR-triggered runs from forks).

assafelovic:
Fixed in `main` via #2074 (commit `fbc03378`).

Your diagnosis was exactly right, and it was the root cause of two shipped crash regressions. `.github/workflows/tests.yml` now runs on every pull request with three jobs:

- `imports` — matrix over 3.12/3.13/3.14, walks every submodule and loads the ASGI app
[...]

PR #1963, opened 2026-07-21
ci: run offline unit pytest suite on pull requests

Bartok9:
## Summary

Adds a GitHub Actions Tests workflow that runs a curated offline unit pytest set on every PR and push to master/main.

## Motivation

Closes the gap reported in #1945: none of the existing workflows invoke pytest, so regressions (including package ImportError bugs that already reached a published release) never block merge.

## Changes

- .github/workflows/tests.yml — Python 3.12, lean deps, fixed list of offline unit modules

## Test plan

- [x] YAML loads
- [x] Workflow targets unittest-style modules that need no API keys
- Author Bartok9

Related: #1945

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,41 @@
+name: Tests
+
+on:
+  pull_request:
+  push:
+    branches: [master, main]
+
+jobs:
+  unit:
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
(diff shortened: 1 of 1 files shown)


assafelovic:
The same defect is now fixed in `main` via #2074 — the version from commit fbc03378 (the tests.yml added there) was the one taken, so this one is redundant rather than unwanted.

PR #1964, opened 2026-07-21
fix(document): tolerate missing metadata source on loaded pages

Bartok9:
## Summary

DocumentLoader no longer KeyErrors when a loaded page omits metadata['source'].

## Motivation

page.metadata['source'] raises and aborts the whole local-doc load gather if any loader omits the key. Use .get with file_path fallback and empty basename when absent.

## Real behavior proof

python3 -m pytest tests/test_document_loader_metadata_source.py -q
1 passed

## Author

Bartok9

### DIFF
--- a/gpt_researcher/document/document.py
+++ b/gpt_researcher/document/document.py
@@ -50,9 +50,14 @@ class DocumentLoader:
         for pages in await asyncio.gather(*tasks):
             for page in pages:
                 if page.page_content:
+                    # Loaders occasionally omit metadata["source"] (custom loaders,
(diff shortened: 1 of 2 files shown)


assafelovic:

It fails one test against the current base. `document.py` has since changed via #1966, which landed, so a rebase may resolve this on its own.

Bartok9:
Thanks — rebased onto current `main` and force-pushed a clean tip.

`document.py` on `main` still indexes `page.metadata['source']` (KeyError if a loader omits it). This PR keeps the `.get` + `file_path` fallback and the regression test. CI should re-run on the new head.

PR #1965, opened 2026-07-21
fix(scraper): guard BeautifulSoup fetch when session is None

Bartok9:
## Summary

BeautifulSoupScraper._fetch returns None when session is None instead of AttributeError on session.get.

## Real behavior proof

python3 -m pytest tests/test_bs_session_none.py -q
1 passed

## Author

Bartok9


assafelovic:

PR #1966, opened 2026-07-21
fix(document): log real extension when loader.load fails

Bartok9:
## Summary

Exception path during DocumentLoader._load_document no longer always says HTML.

## Real behavior proof

pytest tests/test_document_loader_error_message.py -q : 1 passed

Author Bartok9


assafelovic:

PR #1968, opened 2026-07-21
fix(query): move typing imports above _normalize_sub_queries

uxweitt:
`from typing import Any, List, Dict` (and the other module imports) were placed below `_normalize_sub_queries`, whose signature is annotated `parsed: Any` and `-> List[str]`. Python evaluates function annotations at definition time, so importing this module raised `NameError: name 'Any' is not defined`.

Move the imports to the top of the module so the annotations resolve.


assafelovic:

PR #1969, opened 2026-07-22
fix(report): skip available_images entries without url

Bartok9:
## Summary

Harden `generate_report` so partial `available_images` payloads no longer abort report generation with `KeyError`/`TypeError` on `img['url']`.

## Motivation

Image generators and callers can pass `None` rows, non-dicts, or dicts missing/empty `url`. The previous list-comp required `img['url']` for every entry.

## Changes

- Skip non-dict / missing-url image entries when building the AVAILABLE IMAGES block
- Omit the block entirely when no usable URLs remain

## Real behavior proof

```
$ python3 -m pytest tests/test_report_generation_available_images.py -v
2 passed
```

## Test plan

- [x] Unit tests cover mixed invalid/valid image rows and all-invalid omit path
- [x] No change to happy-path embedding when `url` is present

### DIFF
--- a/gpt_researcher/actions/report_generation.py
+++ b/gpt_researcher/actions/report_generation.py
@@ -259,11 +259,28 @@ async def generate_report(
     
     # Add available images instruction if images were pre-generated
     if available_images:
-        images_info = "\n".join([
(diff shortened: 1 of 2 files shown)


assafelovic:

PR #1970, opened 2026-07-22
fix(serper): return empty results on non-dict JSON payloads

Bartok9:
## Summary

Harden `SerperSearch.search` so non-dict JSON payloads (arrays/strings from error or truncated bodies) return `[]` instead of crashing on `.get`.

## Motivation

`json.loads` success only means “valid JSON”, not “object with organic results”. List/string payloads caused `AttributeError` mid-research.

## Changes

- After parse, require `isinstance(search_results, dict)` before reading `organic`
- Existing per-row dict/href guards unchanged

## Real behavior proof

```
$ python3 -m pytest tests/test_serper_non_dict_payload.py -v
3 passed
```

## Test plan

- [x] List payload → `[]`
- [x] String payload → `[]`
- [x] Happy path still normalizes organic rows


Bartok9:
Opened against a stale remote branch tip (fork branchTracking mismatch). Replacing with a clean single-commit PR.

PR #1972, opened 2026-07-22
fix(scraper): keep TavilyExtract content when session is None

Bartok9:
## Summary

Keep Tavily-extracted body text when `session` is `None` or image/title enrichment fails, and harden extract payload shape checks.

## Motivation

`TavilyExtract(link, session=None)` is a normal construction path. The previous code always called `self.session.get` after a successful Tavily extract; the broad `except` then returned empty content, wiping the good extract.

## Changes

- Require dict response; soft-read `results[0].raw_content`
- Only enrich images/title when `session is not None`; failures leave content intact

## Real behavior proof

```
$ python3 -m pytest tests/test_tavily_extract_session_none.py -v
2 passed
```

## Test plan

- [x] `session=None` still returns extracted body
- [x] non-dict extract response → empty triple

### DIFF
--- a/gpt_researcher/scraper/tavily_extract/tavily_extract.py
+++ b/gpt_researcher/scraper/tavily_extract/tavily_extract.py
@@ -37,23 +37,41 @@ class TavilyExtract:
 
         try:
             response = self.tavily_client.extract(urls=self.link)
-            if response['failed_results']:
+            if not isinstance(response, dict):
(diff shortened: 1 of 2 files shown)


assafelovic:
The same defect is now fixed in `main` via #2074 — the version from #1915 was the one taken, so this one is redundant rather than unwanted.

PR #1973, opened 2026-07-22
fix(costs): tolerate None content in estimate_llm_cost

Bartok9:
## Summary

Coerce `None`/non-string arguments in `estimate_llm_cost` so stream and callback paths do not fail inside `tiktoken.encode`.

## Motivation

Cost callbacks sometimes fire with missing text. bare encode raised TypeError.

## Real behavior proof

```
$ python3 -m pytest tests/test_estimate_llm_cost_none.py -v
3 passed
```

## Test plan

- [x] None/None → 0.0
- [x] Mixed None does not raise
- [x] Non-str coerced


assafelovic:

PR #1974, opened 2026-07-23
fix(scraper): tolerate None content before len() checks

Bartok9:
## Summary
- Backend scrapers can return `(None, image_urls, title)` on hard failures.
- `Scraper.extract_data_from_url` called `len(content)` before any None check, raising `TypeError` and risking a failed `asyncio.gather` batch for a single bad URL.
- Coerce `None` content/image_urls/title, then keep the existing short/empty payload path.

## Motivation
One scraper failure returning `None` content should degrade that URL to an empty payload, not throw out of `extract_data_from_url`.

## Test plan
- [x] `pytest tests/test_scraper_none_content.py -v` (3 passed)
- [x] Real behavior: `len(None)` raised `TypeError` before the fix

## Real behavior proof
REPRO TypeError object of type 'NoneType' has no len()
3 passed in test_scraper_none_content.py

### DIFF
--- a/gpt_researcher/scraper/scraper.py
+++ b/gpt_researcher/scraper/scraper.py
@@ -131,25 +131,25 @@ class Scraper:
                         self.worker_pool.executor, scraper.scrape
                     )
 
-                if len(content) < 100:
-                    self.logger.warning(f"Content too short or empty for {link}")
-                    return {
-                        "url": link,
-                        "raw_content": None,
-                        "image_urls": [],
-                        "title": title,
-                    }
+                # Scrapers may return None content on hard failures. Guard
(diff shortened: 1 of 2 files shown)


assafelovic:

It no longer applies: `scraper.py` was substantially changed by #1944 and #1951, which both landed. Needs a rebase to tell whether the guard it adds is still required.

Bartok9:

I rebased/checked this against current `main`: \`extract_data_from_url\` was substantially rewritten (#1944/#1951); \`if not content or len(content) < 100\` already tolerates \`None\`, and exceptions return an empty payload.

Bartok9:

PR #1975, opened 2026-07-23
fix(agent): guard handle_json_error on partial/non-dict payloads

Bartok9:
## Summary
- `handle_json_error` regex path did `json_data[\"server\"], json_data[\"agent_role_prompt\"]` without shape checks.
- Partial LLM JSON (only `server`, non-dict repair results) raised `KeyError` instead of falling back to the default agent.
- Centralize complete-dict validation and skip `json_repair.loads(None)` which thrash-logs inside json_repair.

## Motivation
Agent selection recovery should degrade to Default Agent, not crash the research pipeline.

## Test plan
- [x] `pytest tests/test_agent_creator_json_shape.py -v` (10 passed)
- [x] Demonstrated pre-fix `KeyError` on `{\"server\":\"only-server\"}`

## Real behavior proof
Partial dict path KeyError before fix; 10 tests green after fix.

### DIFF
--- a/gpt_researcher/actions/agent_creator.py
+++ b/gpt_researcher/actions/agent_creator.py
@@ -76,9 +76,10 @@ async def handle_json_error(response: str | None):
         if all parsing attempts fail.
     """
     try:
-        agent_dict = json_repair.loads(response)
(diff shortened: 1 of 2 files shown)


assafelovic:

PR #1976, opened 2026-07-23
fix(markdown): guard extractors against None/empty inputs

Bartok9:
## Summary
- `extract_headers` / `extract_sections` / `table_of_contents` crashed on `None` via `markdown.markdown(None)`.
- `add_references` crashed when report markdown or visited_urls was `None`.
- Return empty lists / safe strings so post-report formatters degrade cleanly.

## Test plan
- [x] `pytest tests/test_markdown_none_inputs.py -v` (5 passed)

## Real behavior proof
Before: `AttributeError: 'NoneType' object has no attribute 'strip'` on extract_headers(None).


assafelovic:

PR #1977, opened 2026-07-23
fix(utils): coerce None token counts in cost helpers

Bartok9:
## Summary
- `format_token_count(None)` raised `TypeError` via f-string formatting.
- `calculate_cost(None, ...)` raised on `prompt_tokens + completion_tokens`.
- Coerce invalid/negative counts to 0 so websocket cost updates with missing usage metadata do not crash.

## Test plan
- [x] `pytest tests/test_utils_token_cost_guard.py -v` (3 passed)

## Real behavior proof
Before: `TypeError: unsupported operand type(s) for +: 'NoneType' and 'int'`.

### DIFF
--- a/gpt_researcher/actions/utils.py
+++ b/gpt_researcher/actions/utils.py
@@ -75,6 +75,23 @@ def calculate_cost(
     Returns:
         float: The calculated cost in USD.
     """
+    try:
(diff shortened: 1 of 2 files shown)


assafelovic:

PR #1983, opened 2026-07-24
fix(server): single LLM-backed report chat route

### LINKED ISSUE #1979: Duplicate route registration: POST /api/reports/{research_id}/chat LLM handler is unreachable
**Found while evaluating gpt-researcher for embedding as a library** (HEAD `5d84d2f`, v0.14.7).

`backend/server/app.py` registers **POST `/api/reports/{research_id}/chat` twice**:

- **line 263** — `add_report_chat_message`: appends the posted message to the JSON `ReportStore`, runs no LLM
- **line 420** — `research_report_chat`: constructs `ChatAgentWithMemory` and actually runs the chat completion

FastAPI/Starlette matches the **first** registered route, so the line-420 LLM handler is unreachable dead code. Any client POSTing to `/api/reports/{id}/chat` expecting an assistant reply just gets its message appended to the store (the Next.js proxy at `frontend/nextjs/app/api/reports/[id]/chat/route.ts` proxies to exactly this path). The only live LLM chat path is `POST /api/chat`, which requires the client to re-send the full report + history each request.

Also duplicated (harmless direction, but confusing dead code): **PUT** `/api/reports/{research_id}` at lines 227 (real, ReportStore-backed) vs 457 ("no database configured" stub), and **DELETE** at 247 vs 463 — the later stubs are unreachable.

**Suggested fix:** collapse to a single POST `/api/reports/{id}/chat` handler that loads the report + history from `ReportStore`, runs `ChatAgentWithMemory`, and persists both the user and assistant messages — which would also remove the need for clients to re-send the report text every turn. Delete the three unreachable duplicates.

Bartok9:
## Summary

- Collapse duplicate POST `/api/reports/{id}/chat` into one ReportStore + LLM handler
- Remove unreachable PUT/DELETE stub routes that shadowed nothing useful after the real ones
- Persist user + assistant messages on chat

## Motivation

Closes #1979.

FastAPI matched the first `add_report_chat_message` registration, so the later `research_report_chat` LLM handler was dead code. Clients expecting an assistant reply only got store-append behavior.

## Verification

- `.venv/bin/python -m pytest tests/backend/test_report_chat_route_uniqueness.py -v` — 2 passed
- Route registry asserts a single POST chat path and single PUT/DELETE for reports/{id}
- Handler mock proves ChatAgentWithMemory runs and messages persist

## Duplicate check

- check-duplicates.sh 1979 — no open PR

### DIFF
--- a/backend/server/app.py
+++ b/backend/server/app.py
@@ -261,27 +261,73 @@ async def get_report_chat(research_id: str):
 
 
 @app.post("/api/reports/{research_id}/chat")
-async def add_report_chat_message(research_id: str, request: Request):
+async def research_report_chat(research_id: str, request: Request):
+    """Chat against a stored report: LLM reply + persist user/assistant messages."""
     report = await report_store.get_report(research_id)
     if report is None:
         raise HTTPException(status_code=404, detail="Report not found")
 
-    message = await request.json()
-    chat_messages = report.get("chatMessages") or []
-    if isinstance(chat_messages, list):
-        chat_messages = [*chat_messages, message]
-    else:
-        chat_messages = [message]
+    try:
+        data = await request.json()
+    except Exception:
+        data = {}
+    if not isinstance(data, dict):
+        data = {}
+
+    # Accept either a bare chat message object or {messages, report, message}
+    incoming_messages = data.get("messages")
+    if not isinstance(incoming_messages, list):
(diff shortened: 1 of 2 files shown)


assafelovic:

FlorianFranzen:

PR #1984, opened 2026-07-24
fix(image_generator): repair LLM JSON in analysis parse

Bartok9:
## Summary

- Parse report-image analysis LLM output with `json_repair` (fenced / trailing commas)
- Skip non-dict suggestion rows; empty/non-str responses return []

## Motivation

`_parse_analysis_response` used regex + `json.loads` only. Common LLM wrappers (markdown fences, trailing commas) dropped all image suggestions silently. Related path `_plan_images` was hardened in #1955; this covers the analysis helper on the same skill.

## Verification

- `.venv/bin/python -m pytest tests/test_image_generator_analysis_json_repair.py -v` — 5 passed
- Covers raw JSON, fenced JSON, trailing-comma repair, non-dict rows, empty input

## Duplicate check

- Open #1955 covers planning JSON recovery, not `_parse_analysis_response` (this PR)

### DIFF
--- a/gpt_researcher/skills/image_generator.py
+++ b/gpt_researcher/skills/image_generator.py
@@ -6,6 +6,7 @@ contextually relevant images for research reports using AI image generation.
 
 import asyncio
 import json
+import json_repair
 import logging
 import re
 from typing import Any, Dict, List, Optional, Tuple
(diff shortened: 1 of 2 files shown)


assafelovic:
The same defect is now fixed in `main` via #2074 — the version from #1955 was the one taken, so this one is redundant rather than unwanted.

PR #1985, opened 2026-07-24
fix(agent): prefer json_repair on choose_agent hot path

Bartok9:
## Summary

- Use `json_repair.loads` first in `choose_agent` before strict `json.loads`
- Keep `handle_json_error` as fallback for unrecoverable payloads

## Motivation

LLM agent-selection responses often arrive fenced or with trailing commas. Strict `json.loads` always threw into the recovery path; repairing on the hot path avoids an unnecessary log/exception berth and matches other skills that already use json_repair.

## Verification

- `.venv/bin/python -m pytest tests/test_create_agent_json_repair_primary.py -v` — 2 passed

## Duplicate check

- No open PR for choose_agent primary-path json_repair (handle_json_error already used repair on failure only)

### DIFF
--- a/gpt_researcher/actions/agent_creator.py
+++ b/gpt_researcher/actions/agent_creator.py
@@ -55,7 +55,16 @@ async def choose_agent(
             **kwargs
         )
 
-        agent_dict = json.loads(response)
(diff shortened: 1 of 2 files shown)


assafelovic:

PR #1988, opened 2026-07-25
fix(actions): import typing before Any annotations

### LINKED ISSUE #1981: v0.16.0 PyPI wheel fails on import: NameError: name 'Any' is not defined (actions/query_processing.py)
**Repro** (clean venv, Python 3.12):

```
pip install gpt-researcher==0.16.0
python -c "from gpt_researcher import GPTResearcher"
```

```
File ".../gpt_researcher/actions/query_processing.py", line 6, in <module>
    def _normalize_sub_queries(parsed: Any, fallback_query: str) -> List[str]:
```

`query_processing.py` uses `Any` (and `List`) in annotations evaluated at def-time without importing them from `typing`. Every import of the package fails, so the 0.16.0 wheel is unusable as a library.

Fix: add `from typing import Any, List` (or `from __future__ import annotations`) to `gpt_researcher/actions/query_processing.py`.

Found while evaluating 0.16.0 as an upgrade target from 0.14.7 (which imports fine).

Bartok9:
## Summary
`query_processing.py` used `Any`/`List` in annotations before importing them, so `from gpt_researcher import GPTResearcher` raises `NameError` on 0.16.0.

## Changes
- Add `from __future__ import annotations` and move `typing` imports to the top
- Regression test for import-order / annotation safety

Closes #1981

## Test plan
- `pytest tests/test_query_processing_typing_import.py -q` (2 passed)

### DIFF
--- a/gpt_researcher/actions/query_processing.py
+++ b/gpt_researcher/actions/query_processing.py
@@ -1,4 +1,7 @@
+from __future__ import annotations
+
 import json_repair
+from typing import Any, Dict, List
 
 from gpt_researcher.llm_provider.generic.base import ReasoningEfforts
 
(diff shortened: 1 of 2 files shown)


assafelovic:

PR #1989, opened 2026-07-25
fix(costs): price Anthropic prompt-cache tokens

### LINKED ISSUE #1986: Anthropic prompt-cache tokens (cache_creation_input_tokens / cache_read_input_tokens) are silently priced at $0
**Describe the bug**

`gpt_researcher/utils/costs.py`'s Anthropic cost tracking never reads `cache_creation_input_tokens` or `cache_read_input_tokens` from the usage dict, so any research run that triggers Anthropic prompt caching (large repeated system prompts, which gpt-researcher sends on nearly every LLM call) silently prices the cached tokens at $0 instead of at Anthropic's real cache-write (1.25x input rate) / cache-read (0.1x input rate) billing.

`_extract_anthropic_usage` (costs.py:106-131) only extracts `input_tokens` and `output_tokens` from `usage`:

```python
if usage:
    input_tokens = usage.get("input_tokens")
    output_tokens = usage.get("output_tokens")
    if input_tokens is not None and output_tokens is not None:
        return {
            "input_tokens": int(input_tokens),
            "output_tokens": int(output_tokens),
        }
```

`calculate_anthropic_cost` (costs.py:159-182) then only prices those two fields. Anthropic's Messages API reports `cache_creation_input_tokens` and `cache_read_input_tokens` as separate, real, billed fields alongside `input_tokens`, and neither is read anywhere in this file.

**To Reproduce**

Load `costs.py` directly and call `calculate_llm_cost` with an Anthropic `response_metadata["usage"]` that includes a cache write, once with the cache fields present and once without:

```python
import importlib.util
spec = importlib.util.spec_from_file_location("costs", "gpt_researcher/utils/costs.py")
costs = importlib.util.module_from_spec(spec)
spec.loader.exec_module(costs)

usage_with_cache = {"input_tokens": 50, "cache_creation_input_tokens": 4000, "cache_read_input_tokens": 0, "output_tokens": 300}
usage_without_cache_fields = {"input_tokens": 50, "output_tokens": 300}

for usage in (usage_with_cache, usage_without_cache_fields):
    cost = costs.calculate_llm_cost(
        llm_provider="anthropic",
        model="claude-sonnet-4-5",
        input_content="", output_content="",
        response_metadata={"model": "claude-sonnet-4-5", "usage": usage},
    )
    print(usage, "->", cost)
```

Output (ran in a clean `python:3.11-slim` container, HEAD `5d84d2f5553e70a2765a8ff3a0d2672d60437ce8`):

```
{'input_tokens': 50, 'cache_creation_input_tokens': 4000, 'cache_read_input_tokens': 0, 'output_tokens': 300} -> 0.00465
{'input_tokens': 50, 'output_tokens': 300} -> 0.00465
```

**Expected behavior**

The 4000 `cache_creation_input_tokens` are real, billed tokens (at 1.25x the base input rate for claude-sonnet-4-5, that's an extra ~$0.015 on top of the $0.00465 above). Adding or removing them from the usage dict should change the reported cost. Instead the two calls return identical costs: the cache tokens are priced at $0.

**Additional context**

`tests/test_costs.py` has no test case involving either cache field, and `git log -G` on both field names over `costs.py`'s whole history returns no commits, so this looks like a straight gap rather than a deliberate simplification. Happy to send a PR if useful; the fix looks like adding both fields to `_extract_anthropic_usage`'s returned dict and pricing them at the write/read multipliers in `calculate_anthropic_cost`.

Bartok9:
## Summary
Anthropic usage includes `cache_creation_input_tokens` / `cache_read_input_tokens`, but cost helpers only priced `input_tokens`+`output_tokens`, so cache traffic was billed at $0.

## Changes
- Extract cache fields in `_extract_anthropic_usage`
- Price cache write at 1.25× input and cache read at 0.1× input in `calculate_anthropic_cost`
- Regression tests

Closes #1986

## Test plan
- `pytest tests/test_anthropic_cache_costs.py -q` (2 passed)

### DIFF
--- a/gpt_researcher/utils/costs.py
+++ b/gpt_researcher/utils/costs.py
@@ -103,32 +103,52 @@ def _resolve_anthropic_model_name(
     ).lower()
 
 
+def _coerce_token_count(value: Any) -> int:
+    if value is None:
+        return 0
+    try:
+        return int(value)
+    except (TypeError, ValueError):
+        return 0
+
+
 def _extract_anthropic_usage(
     response_metadata: Mapping[str, Any] | None = None,
     usage_metadata: Mapping[str, Any] | Any | None = None,
 ) -> dict[str, int] | None:
+    """Extract Anthropic usage including prompt-cache token fields.
+
+    Anthropic Messages usage may include:
+    - input_tokens / output_tokens
+    - cache_creation_input_tokens (billed ~1.25x input)
+    - cache_read_input_tokens (billed ~0.1x input)
+    """
     metadata = _mapping_to_dict(response_metadata)
     usage = _mapping_to_dict(metadata.get("usage"))
 
+    def _from_usage(usage_dict: dict[str, Any]) -> dict[str, int] | None:
+        input_tokens = usage_dict.get("input_tokens")
+        output_tokens = usage_dict.get("output_tokens")
+        if input_tokens is None or output_tokens is None:
+            return None
+        return {
+            "input_tokens": int(input_tokens),
+            "output_tokens": int(output_tokens),
+            "cache_creation_input_tokens": _coerce_token_count(
+                usage_dict.get("cache_creation_input_tokens")
+            ),
+            "cache_read_input_tokens": _coerce_token_count(
+                usage_dict.get("cache_read_input_tokens")
+            ),
+        }
+
     if usage:
-        input_tokens = usage.get("input_tokens")
-        output_tokens = usage.get("output_tokens")
-        if input_tokens is not None and output_tokens is not None:
-            return {
-                "input_tokens": int(input_tokens),
-                "output_tokens": int(output_tokens),
-            }
+        parsed = _from_usage(usage)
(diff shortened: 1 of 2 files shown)


assafelovic:

FlorianFranzen:

Bartok9:
Thank you @FlorianFranzen — you are right, and I appreciate the heads-up.

@assafelovic — when #1989 landed via #2074 in `39215c5b` (*fix: land 15 issue-backed community fixes*), the squash listed `#1989` in the body but the only `Co-Authored-By` trailer is for Claude Opus 5. Could you please add a human trailer when you next touch that history (or a small follow-up commit) so the Anthropic cache-token pricing fix is attributed?

```
Co-Authored-By: Bartok9 <259807879+Bartok9@users.noreply.github.com>
```

No rush on a rewrite of main — just asking for the byline to match the consolidation note. Grateful the fix shipped.

PR #1990, opened 2026-07-25
fix(scraper): tolerate None in parse_dimension

Bartok9:
## Summary
`parse_dimension` called `value.lower()` on HTML dim attrs; `img.get('width')` is often `None` and crashed image scoring.

## Changes
- Guard None/non-str inputs
- Unit tests

## Test plan
- `pytest tests/test_parse_dimension_none.py -q`


assafelovic:

PR #1994, opened 2026-07-26
fix(semantic_scholar): guard openAccessPdf and payload shapes

Bartok9:
## Summary

- Guard non-dict JSON/`data` rows and string/null `openAccessPdf` shapes.
- Preserve API sort casing (`citationCount` / `publicationDate`).

## Motivation

`openAccessPdf.get` crashed when the field was a bare URL string; lowering sort broke valid Semantic Scholar enum values.

## Verification

- `.venv/bin/python -m pytest tests/retrievers/test_semantic_scholar_guards.py -q` — 5 passed

## Duplicate check

- Prior #1897 closed; main still has `.get` on openAccessPdf without shape guards.

### DIFF
--- a/gpt_researcher/retrievers/semantic_scholar/semantic_scholar.py
+++ b/gpt_researcher/retrievers/semantic_scholar/semantic_scholar.py
@@ -20,10 +20,7 @@ class SemanticScholarSearch:
         """
         self.query = query
         assert sort in self.VALID_SORT_CRITERIA, "Invalid sort criterion"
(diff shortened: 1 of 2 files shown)


assafelovic:

It fails one test against the current base.

Bartok9:

I rebased/checked this against current `main`: Semantic Scholar \`openAccessPdf\` / payload shape guards are already on \`main\`.

Bartok9:

PR #1995, opened 2026-07-26
fix(xquik): skip non-dict tweets and empty ids

Bartok9:
## Summary

- Guard Xquik JSON root/`tweets` list and skip non-dict/id-less rows.
- Tolerate alternate author/text field names.

## Verification

- `.venv/bin/python -m pytest tests/retrievers/test_xquik_tweet_shapes.py -q` — 3 passed

## Duplicate check

- No open PR on xquik tweet shape guards.

### DIFF
--- a/gpt_researcher/retrievers/xquik/xquik.py
+++ b/gpt_researcher/retrievers/xquik/xquik.py
@@ -68,30 +68,46 @@ class XquikSearch:
         with urllib.request.urlopen(req, timeout=15) as resp:
             data = json.loads(resp.read().decode("utf-8"))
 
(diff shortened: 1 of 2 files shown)


assafelovic:

It fails one test against the current base. `xquik.py` was since fixed via #1937, which landed — this may now be redundant, but that's worth your check rather than my guess.

Bartok9:

I rebased/checked this against current `main`: xquik non-dict tweet / empty-id guards are already on \`main\` (via #1937 / follow-ups).

Bartok9:

PR #1996, opened 2026-07-26
fix(chat): enable report RAG path and fix retriever kwargs

### LINKED ISSUE #1980: ChatAgentWithMemory: vector-store RAG path is dead code (`and False`) and self.retriever is never consulted
**Found while evaluating gpt-researcher for embedding as a library** (HEAD `5d84d2f`, v0.14.7).

`backend/chat/chat.py` contains a vector-store retrieval path for chat that can never run, and would not be used even if it did:

1. **Hard-disabled guard** — line 79:
   ```python
   if not self.vector_store and False:
       self._setup_vector_store()
   ```
   `_setup_vector_store` / `_process_document` (lines 82–112) are unreachable.

2. **Retriever never consulted** — `self.retriever` is assigned (lines 67, 101) but never read anywhere in the class. `chat()` unconditionally injects the **entire report** into the system prompt every turn (`Report: {self.report}`, ~line 216), and `process_chat_completion` never touches the retriever. So chat is context-window-bound: fine for typical reports, but long `detailed_report`/`deep` outputs plus long histories eventually blow the window, and every turn pays the full-report token cost.

3. **Two latent bugs inside the dead path** (would surface if the guard were removed):
   - line 91: `cfg = Config()` builds a fresh default config, ignoring the `config_path` the agent was constructed with — embedding provider/model overrides are dropped.
   - line 101: `self.vector_store.as_retriever(k=4)` — `k` is not a supported kwarg; should be `search_kwargs={"k": 4}` (it only behaves as intended today because the default k is 4).

**Suggested direction if RAG chat is wanted:** remove the guard, fix the two bugs, add a `_retrieve_context(user_message)` helper (`await self.retriever.ainvoke(...)`), and build the system prompt from retrieved chunks (falling back to full-report injection for short reports). Accepting a caller-supplied `retriever=` (in addition to the existing `vector_store=` ctor arg) would let embedders plug their own retrieval backends without local embedding.

Happy to provide more detail from our evaluation if useful. Related routing issue: #1979.

Bartok9:
## Summary

- Remove `and False` guard so report chat can build an in-memory vector store.
- Use agent `Config` (not a fresh default) for embeddings; `search_kwargs={"k": 4}`.
- Retrieve relevant report chunks for system context; fall back to full report.
- Treat `tavily` as optional import for environments without the package.

## Motivation

Closes #1980.

## Verification

- `PYTHONPATH=. .venv/bin/python -m pytest tests/chat/test_chat_agent_rag_path.py -q` — 3 passed

## Duplicate check

- check-duplicates / gh: no open fix PR for #1980 (only unrelated dependabot hit).

### DIFF
--- a/backend/chat/chat.py
+++ b/backend/chat/chat.py
@@ -11,7 +11,10 @@ from gpt_researcher.memory import Memory
 from gpt_researcher.config.config import Config
 from gpt_researcher.utils.llm import create_chat_completion
 from gpt_researcher.utils.tools import create_chat_completion_with_tools, create_search_tool
-from tavily import TavilyClient
+try:
+    from tavily import TavilyClient
+except ImportError:  # optional dependency for chat web search
+    TavilyClient = None
 from datetime import datetime
 
 # Setup logging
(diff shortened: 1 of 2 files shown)


AmirF194 (commented):
The dead `and False` branch and the retriever kwargs compatibility shim are both real bugs, good fix.

One thing worth flagging: enabling `_setup_vector_store()` (previously unreachable) means every real call site now pays for embeddings at chat-construction time, and it is not optional. All three call sites construct `ChatAgentWithMemory(report=..., config_path="default")` with no `vector_store` (`backend/server/app.py:390`, `backend/server/app.py:430`, `backend/server/server_utils.py:220`), so any non-empty report goes through `Memory(cfg.embedding_provider, cfg.embedding_model, **cfg.embedding_kwargs).get_embeddings()`. The default config (`EMBEDDING = "openai:text-embedding-3-small"`) resolves to `OpenAIEmbeddings(model=...)`, which raises at construction, not lazily, when `OPENAI_API_KEY` is unset. Verified in a clean `python:3.11-slim` container with only `langchain-openai` installed:

```
$ python3 -c "from langchain_openai import OpenAIEmbeddings; OpenAIEmbeddings(model='text-embedding-3-small')"
openai.OpenAIError: Missing credentials. Please pass an `api_key`, `workload_identity`, `admin_api_key`, or set the `OPENAI_API_KEY` or `OPENAI_ADMIN_KEY` environment variable.
```

All three call sites wrap the constructor in a broad `try/except Exception`, so this will not crash the process, but it does mean chat breaks for every message on any deployment whose LLM provider is not OpenAI and that has not separately set `OPENAI_API_KEY` for embeddings. Before this PR that branch was dead code, so chat always worked off the full report regardless of embedding config. Might be worth catching the embedding setup failure inside `_setup_vector_store` and falling back to no-RAG (full report) mode, the same way `_retrieve_context` already falls back when retrieval itself fails.

Non-blocking, the core fix looks right.

Bartok9:
Good catch, thanks @AmirF194 — that eager-embedding cost is real. Enabling `_setup_vector_store()` meant `Memory(...).get_embeddings()` now runs at chat construction, and `OpenAIEmbeddings` raises at init (not lazily) when `OPENAI_API_KEY` is unset, so non-OpenAI deployments would break on every message.

Fixed in 48797a3b: `_setup_vector_store` now wraps the embedding + vector-store construction in a try/except and falls back to no-RAG / full-report mode on failure (embedding/vector_store/retriever set to None), mirroring the existing `_retrieve_context` fallback. Added a test (`test_setup_vector_store_falls_back_when_embeddings_fail`) asserting the fallback and that `_retrieve_context` then serves the full report. Behavior for deployments without embeddings is now the same as before this PR.

assafelovic:

FlorianFranzen:

PR #1998, opened 2026-07-27
fix(bing): skip non-dict webPages.value rows

Bartok9:
## Summary
Skip non-dict (and non-list) Bing webPages.value rows so one bad entry cannot abort search.

## Motivation
Bing occasionally returns odd payload shapes. Calling .get on a string/null row raised and emptied the whole run.

## Verification
```
python3 -m pytest tests/test_bing_retriever_malformed.py -q
4 passed
```

### DIFF
--- a/gpt_researcher/retrievers/bing/bing.py
+++ b/gpt_researcher/retrievers/bing/bing.py
@@ -70,26 +70,33 @@ class BingSearch():
             return []
         try:
             search_results = json.loads(resp.text)
-            results = search_results.get("webPages", {}).get("value", [])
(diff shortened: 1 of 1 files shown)


assafelovic:

It fails one test against the current base. `bing.py` was since fixed via #1931, which landed — this may now be redundant.

Bartok9:

I rebased/checked this against current `main`: Bing non-dict \`webPages.value\` guards are already on \`main\` (via #1931 / follow-ups).

Bartok9:

PR #1999, opened 2026-07-27
fix(brave): skip non-dict web.results rows

Bartok9:
## Summary
Skip non-dict Brave web.results rows and reject non-list payloads.

## Motivation
Calling .get on a non-dict result aborts the whole Brave search.

## Verification
```
python3 -m pytest tests/test_brave_retriever_malformed.py -q
2 passed
```


assafelovic:

PR #2000, opened 2026-07-27
fix(bocha): skip non-dict webPages.value rows

Bartok9:
## Summary
Skip non-dict BoCha webPages.value rows and reject non-list payloads.

## Motivation
Calling .get on a non-dict BoCha hit aborted the search path.

## Verification
```
python3 -m pytest tests/test_bocha_retriever_malformed.py -q
2 passed
```

### DIFF
--- a/gpt_researcher/retrievers/bocha/bocha.py
+++ b/gpt_researcher/retrievers/bocha/bocha.py
@@ -40,32 +40,25 @@ class BoChaSearch():
             "count": max_results
         }
 
-        try:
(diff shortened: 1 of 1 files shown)


assafelovic:

It fails two tests against the current base. `bocha.py` was since fixed via #1921, which landed — this may now be redundant.

Bartok9:

I rebased/checked this against current `main`: BoCha non-dict / non-list \`webPages.value\` guards are already on \`main\` (via #1921 / follow-ups).

Bartok9:

PR #2001, opened 2026-07-27
fix(searchapi): guard organic_results shape and rows

Bartok9:
## Summary
Harden SearchApi organic_results parsing: missing key, non-list, non-dict rows, missing links.

## Motivation
search_results["organic_results"] and result["link"] raised on common empty/error payloads and aborted search.

## Verification
```
python3 -m pytest tests/test_searchapi_retriever_malformed.py -q
3 passed
```


assafelovic:

PR #2002, opened 2026-07-28
fix(scraper): use session for PyMuPDF PDF downloads

### LINKED ISSUE #1847: PyMuPDFScraper ignores the User-Agent session — all PDF downloads go out as python-requests, SEC EDGAR returns 403

Bartok9:
## Summary
Use the provided `requests.Session` (User-Agent) for `PyMuPDFScraper` HTTP PDF downloads instead of bare `requests.get`.

## Motivation
Closes #1847. SEC EDGAR and similar hosts 403 default `python-requests/*` UAs while HTML scrapes with the configured session work.

## Real behavior proof
```
python3 -m pytest tests/test_pymupdf_session_ua.py -q
# 1 passed
```

## Test plan
- [x] unit: session.get used; bare requests.get not called

### DIFF
--- a/gpt_researcher/scraper/pymupdf/pymupdf.py
+++ b/gpt_researcher/scraper/pymupdf/pymupdf.py
@@ -41,15 +41,18 @@ class PyMuPDFScraper:
         """
         try:
             if self.is_url():
+                # Prefer caller-provided session so User-Agent / headers apply
+                # (bare requests.get uses python-requests/* and 403s on SEC EDGAR).
+                http = self.session if self.session is not None else requests
                 try:
-                    response = requests.get(self.link, timeout=(5, 30), stream=True)
+                    response = http.get(self.link, timeout=(5, 30), stream=True)
                     response.raise_for_status()
                 except requests.exceptions.SSLError:
                     import logging
                     logging.getLogger(__name__).warning(
                         f"SSL verification failed for {self.link}, retrying without verification"
                     )
-                    response = requests.get(self.link, timeout=(5, 30), stream=True, verify=False)
+                    response = http.get(self.link, timeout=(5, 30), stream=True, verify=False)
                     response.raise_for_status()
 
                 with tempfile.NamedTemporaryFile(delete=False, suffix=".pdf") as temp_file:
--- a/tests/test_pymupdf_session_ua.py
+++ b/tests/test_pymupdf_session_ua.py
@@ -1,0 +1,75 @@
+"""PyMuPDFScraper must use the provided session for HTTP PDF downloads (#1847)."""
+
+from __future__ import annotations
+
+import importlib.util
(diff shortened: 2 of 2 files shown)


assafelovic:
The same defect is now fixed in `main` via #2074 — the version from #1849 was the one taken, so this one is redundant rather than unwanted.

PR #2003, opened 2026-07-28
fix(vector_store): skip non-dict docs and missing content

Bartok9:
## Summary
`VectorStoreWrapper._create_langchain_documents` skips non-dict rows and null `raw_content`, and tolerates missing `url`.

## Motivation
Bad crawler rows (None / bare strings / partial dicts) used to raise KeyError mid-load and drop the whole batch.

## Real behavior proof
```
python3 -m pytest tests/test_vector_store_doc_guards.py -q
# 1 passed
```

## Test plan
- [x] mixed bad/good document list


assafelovic:

PR #2004, opened 2026-07-28
fix(scraper): harden image class parsing in get_relevant_images

Bartok9:
## Summary
`get_relevant_images` returns `[]` for `None` soup and normalizes non-list `class` attributes before membership checks.

## Motivation
Odd `class` values (string / non-sequence) or a missing soup could error out of image scoring and drop all images for the page.

## Real behavior proof
```
python3 -m pytest tests/test_scraper_images_class_guard.py -q
# 3 passed
```

## Test plan
- [x] None soup; string class; non-sequence class


assafelovic:

PR #2005, opened 2026-07-28
fix(scraper): skip null docs in WebBaseLoaderScraper

Bartok9:
## Summary
`WebBaseLoaderScraper` skips `None` loader docs and entries without `page_content` instead of throwing mid-concat.

## Motivation
LangChain loaders can return sparse document lists; one bad entry emptied the whole scrape via exception path.

## Real behavior proof
```
python3 -m pytest tests/test_web_base_loader_docs_guard.py -q
# 1 passed
```

## Test plan
- [x] mixed None/missing/valid page_content


assafelovic:

PR #2006, opened 2026-07-29
fix(serpapi): guard organic_results shape and rows

Bartok9:
## Summary
SerpApi `organic_results` is assumed to be a list of dicts. Non-list payloads and non-dict rows raised during normalize and dropped the full result set.

## Motivation
Malformed or error-shaped API JSON should not crash search.

## Verification
```
python -m pytest tests/retrievers/test_serpapi_organic_guards.py -q
# 2 passed
```

Real proof: non-list `organic_results` → `[]`; mixed rows keep only valid links (YouTube skipped).


assafelovic:
The same defect is now fixed in `main` via #2074 — the version from #1933 was the one taken, so this one is redundant rather than unwanted.

PR #2008, opened 2026-07-29
fix(crw): skip non-dict data rows from fastCRW

Bartok9:
## Summary
fastCRW `data` list entries are assumed dicts; non-dict rows crashed result normalization.

## Verification
```
python -m pytest tests/retrievers/test_crw_source_guards.py -q
# 2 passed
```


assafelovic:
The same defect is now fixed in `main` via #2074 — the version from #1923 was the one taken, so this one is redundant rather than unwanted.

PR #2009, opened 2026-07-29
fix(getxapi): guard tweet list and author shapes

Bartok9:
## Summary
GetXAPI tweet arrays and `author` objects are not always well-typed; bad rows crashed normalization.

## Verification
```
python -m pytest tests/retrievers/test_getxapi_tweet_guards.py -q
# 2 passed
```


assafelovic:
The same defect is now fixed in `main` via #2074 — the version from #1918 was the one taken, so this one is redundant rather than unwanted.

Bartok9:
Thanks @assafelovic — understood, closing as superseded by the #2074 stack is fine. Appreciate you taking the fix into main.

PR #2010, opened 2026-07-30
fix(mcp): skip non-dict server configs in client conversion

Bartok9:
## Summary
Skip non-dict entries while converting MCP server configs to langchain format.

## Motivation
A single bad list item did `config.get` and crashed conversion.

## Verification
```
.venv/bin/python -m pytest tests/test_mcp_client_non_dict_config.py -q
1 passed
```


assafelovic:

Bartok9:
Thanks @assafelovic — glad this landed via #2074 (`bf6f1b3f`) with co-author credit. Appreciate the actions/skills/MCP consolidation and multi-version suite run.

PR #2011, opened 2026-07-30
fix(scraper): guard filter_urls against non-str URLs

Bartok9:
## Summary
Skip non-string URL entries and tolerate bad `excluded_domains` shapes in `filter_urls`.

## Motivation
Null/int URLs TypeError’d the domain membership check; non-list excluded_domains crashed the same path.

## Verification
```
.venv/bin/python -m pytest tests/test_filter_urls_guards.py -q
3 passed
```


assafelovic:

Bartok9:
Thanks @assafelovic — glad this landed via #2074 (`bf6f1b3f`) with co-author credit. Appreciate stacking the scraper/MCP guards and the multi-version suite run.

PR #2012, opened 2026-07-30
fix(utils): coerce rate_limit_delay in GlobalRateLimiter.configure

Bartok9:
## Summary
Coerce/validate `rate_limit_delay` in `GlobalRateLimiter.configure`.

## Motivation
String/None env values break float comparisons in `wait_if_needed`.

## Verification
```
.venv/bin/python -m pytest tests/test_rate_limiter_configure.py -q
4 passed
```


assafelovic:

Bartok9:
Thanks @assafelovic — glad this landed via #2074 (`3de74650`) with co-author credit. Appreciate stacking the utils/config guards and verifying across 3.12–3.14.

PR #2013, opened 2026-07-30
fix(pubmed): include nested tags in article title text

Bartok9:
## Summary
Collect PubMed article titles with `itertext()` so nested formatting tags are kept.

## Motivation
`.text` only returns the first text node and dropped content inside italic/bold tags.

## Verification
```
.venv/bin/python -m pytest tests/test_pubmed_title_itertext.py -q
1 passed
```


assafelovic:

Bartok9:
Thanks @assafelovic — glad this landed via #2074 (`9140bde5`) with co-author credit. Appreciate the retriever consolidation and multi-version suite run.

PR #2014, opened 2026-07-30
fix(query-processing): import typing names before use

sxwxs:
Move typing and module imports above _normalize_sub_queries so Python 3.11 can evaluate its annotations during module import without raising NameError.


assafelovic:

PR #1867, opened 2026-07-07
[WIP] feat(retrievers): add nimble web search retriever

wildcard:
## Hosted playground

- **Protected playground:** [Open the exact-revision playground](https://gpt-researcher-nimble-playground.kadosh.workers.dev) (access credentials are shared privately).
- **Live recording:** [Watch the authenticated Nimble research run](https://github.com/wildcard/gpt-researcher/blob/2a603a112b4aab75309a2f3d3b4b952c9bf6de0d/gpt-researcher-nimble-hosted-a0f2.mp4).
- **Revision:** `a0f2e1d6700f62447d61742fbc259dc2e3ebad45`.
- **Proof:** the recording shows the decision-oriented web query, live research progress, retrieved source URLs, and the completed cited report. The run used `RETRIEVER=nimble` with a real Nimble credential.
## Summary

- Adds a **`nimble`** retriever so research can run on the [Nimble Search API](https://docs.nimbleway.com/nimble-sdk/web-tools/search) with `RETRIEVER=nimble` and `NIMBLE_API_KEY`.
- `NimbleSearch` mirrors the existing `tavily` and `brave` retrievers: raw `requests` call, `{"href", "body", "title"}` result dicts, `[]` on failure — no new dependencies.
- Wires in the usual two spots — a `"nimble"` case in the `get_retriever` factory and a `NimbleSearch` export from `gpt_researcher.retrievers` — plus a one-line entry in the search-engines docs.
- Ships mocked unit tests, a live-API smoke, and a full end-to-end report run (`RETRIEVER=nimble` → a cited report from Nimble-retrieved sources).

## Demo

**Date:** 2026-08-26 (post-rebase onto current main, HEAD `a0f2e1d6`)

**Query shown:** What new real-time web search and data-access tools for AI agents launched in the last month?

**Proof:** playground used `RETRIEVER=nimble`; logs showed `Active retrievers: ['NimbleSearch']`.

This is a live playground capture of the rebased Nimble retriever, not the July 7 capture.

gptr-nimble-demo

<details>
<summary>Stills — start / retrieving / report / sources</summary>

| Start | Retrieving |
| --- | --- |
| start | retrieving |

| Report | Sources |
| --- | --- |
| report | sources |

</details>

## Motivation

[...]

PR #2015, opened 2026-07-30
feat(retrievers): add Microsoft Web IQ support

sxwxs:
Add an opt-in Web IQ Web Search v3 retriever with per-request API key injection, domain filtering, configuration documentation, malformed-response guards, and mocked unit tests.

Follow the official document: https://webiq.microsoft.ai/documentation/ and tested using real API.


assafelovic:
Reviewed — this is the most thorough of the retriever PRs (10 files, tests, docs, `.env.example`, config reference, and the most error handling of the set). It **no longer applies to `main`** after #2074 and needs a rebase.

One thing worth confirming in the description when you rebase: the endpoint is `https://api.microsoft.ai`. Since the PR is titled *Microsoft Web IQ*, it would help to state plainly whether this is an official Microsoft service and where its docs live — a first-party API is a different maintenance proposition from a third-party one.

---

PR #2017, opened 2026-07-31
fix(scraper): guard filter_urls against non-str URLs

Bartok9:
## Summary
`filter_urls` crashed with TypeError on None/int URL rows or bad `excluded_domains`; skips unusable entries and always returns a list.

## Test plan
- `pytest tests/test_filter_urls_guards.py -v` (4 passed)


assafelovic:
The same defect is now fixed in `main` via #2074 — the version from #2011 was the one taken, so this one is redundant rather than unwanted.

Bartok9:
Thanks @assafelovic — understood, closing as superseded by the #2074 stack is fine. Appreciate you taking the fix into main.

PR #2018, opened 2026-07-31
fix(reviewer): tolerate missing/None guidelines list

Bartok9:
## Summary
`ReviewerAgent.review_draft` crashed when `task["guidelines"]` was None/missing (`TypeError` on join). Coerce to a list and stringify entries.

## Test plan
- `pytest tests/test_reviewer_guidelines_guard.py -v` (2 passed)

### DIFF
--- a/multi_agents/agents/reviewer.py
+++ b/multi_agents/agents/reviewer.py
@@ -18,8 +18,11 @@ class ReviewerAgent:
         :param draft_state:
         :return:
         """
-        task = draft_state.get("task")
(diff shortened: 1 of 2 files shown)


assafelovic:
Landed in `main` via #2074 — commit `705d4af8` (*fix(multi_agents): tolerate missing guidelines and null research data*).

Bartok9:
Thanks @assafelovic — glad this landed via #2074 (`705d4af8`) with co-author credit. Appreciate the multi-agent consolidation and 3.12–3.14 suite run.

PR #2020, opened 2026-07-31
fix(pubmed): guard esearch idlist shape before iterating

Bartok9:
## Summary
eutils `idlist` is not always a list; harden `_search_articles` so non-list / mixed entries return a clean `list[str]` (or empty).

## Test plan
- `pytest tests/test_pubmed_idlist_shape.py -v` (2 passed)


assafelovic:

It fails seven tests against the current base. `pubmed_central.py` was since fixed via #1940, which landed — this may now be redundant.

Bartok9:

I rebased/checked this against current `main`: the pubmed \`idlist\` list/shape guards are already present in \`pubmed_central.py\` (via #1940 / follow-ups).

Bartok9:

PR #2026, opened 2026-08-04
fix(chat): avoid duplicate tool searches

RerankerGuo:
## Summary

Reuse search metadata produced by the chat tool execution instead of issuing the
same Tavily query a second time.

## Problem

`create_chat_completion_with_tools()` already executes `search_tool`, which
calls `ChatAgentWithMemory.quick_search()`. When processing the returned tool
metadata, `process_chat_completion()` called `quick_search(query)` again solely
to rebuild frontend source metadata.

One LLM tool call therefore produced two identical Tavily API requests, adding
latency and potentially doubling search usage/cost.

## Changes

- Store search metadata by query during the original tool execution.
- Reset the per-completion metadata map before each chat completion.
- Read the cached metadata while formatting frontend tool metadata.
- Record query-specific error metadata when Tavily is unavailable or fails.

## Verification

Before:

```text
Expected one Tavily call; observed two identical calls.
```

After:

```text
tests/backend/test_chat_search_metadata.py ..  2 passed
```

The tests verify:

- one Tavily request per LLM search tool call while preserving source metadata;
- the disabled-search path records query-specific error metadata.

local test process supplied those names without modifying that module.

## Duplicate Check

Open PR #1996 also changes `backend/chat/chat.py`, but its scope is report RAG
and vector-store initialization. It does not change quick-search execution or
tool metadata processing.

## Impact

- **Breaking change:** No
- **Affected area:** Report chat web-search tool
- **API usage:** Removes one redundant request per search tool call

## Checklist

- [x] Search source metadata remains available to the frontend.
- [x] Disabled-search behavior remains covered.
- [x] Open PR titles, bodies, diffs, and changed files were checked.

### DIFF
--- a/backend/chat/chat.py
+++ b/backend/chat/chat.py
@@ -113,6 +113,9 @@ class ChatAgentWithMemory:
 
     def quick_search(self, query):
         """Perform a web search for current information using Tavily"""
+        if not hasattr(self, "search_metadata_by_query"):
+            self.search_metadata_by_query = {}
+
         try:
             # Check if Tavily client is available
             if self.tavily_client is None:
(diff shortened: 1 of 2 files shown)


RerankerGuo:

PR #2028, opened 2026-08-04
fix(azure): clean unused download directories

RerankerGuo:
## Summary

Remove Azure document download directories when loading fails, is cancelled, or
produces no files.

## Problem

`AzureDocumentLoader.load()` creates a directory with `tempfile.mkdtemp()`.
Successful downloads must remain available for the downstream `DocumentLoader`,
but two terminal paths had no consumer and still leaked the directory:

- path validation or download failures;
- an empty blob container.

Repeated failed/empty loads therefore accumulated unused directories under the
system temporary location.

## Changes

- Remove the download directory before re-raising any exception, including
  cancellation.
- Remove the directory when no blob files were downloaded.
- Preserve the existing successful-load behavior and returned file paths.
- Extend the existing Azure loader tests with explicit cleanup assertions.

## Verification

Before:

```text
2 failed, 1 passed
```

Both the unsafe-blob and empty-container tests observed a surviving directory.

After:

```text
tests/test_azure_document_loader.py ...  3 passed
```

No Azure account or network calls are required; the tests use the existing fake
container and blob clients.

## Impact

- **Breaking change:** No
- **Successful downloads:** Preserved for downstream parsing
- **Affected paths:** Failed, cancelled, and empty Azure loads

## Checklist

- [x] The cleanup assertions fail before and pass after the fix.
- [x] Successful nested blob downloads remain covered.
- [x] Unsafe path validation remains covered.

### DIFF
--- a/gpt_researcher/document/azure_document_loader.py
+++ b/gpt_researcher/document/azure_document_loader.py
@@ -1,6 +1,8 @@
-from azure.storage.blob import BlobServiceClient
+import shutil
 import tempfile
 from pathlib import Path, PurePosixPath
+
+from azure.storage.blob import BlobServiceClient
 
 
 class AzureDocumentLoader:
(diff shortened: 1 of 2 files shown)


RerankerGuo:

PR #2029, opened 2026-08-04
fix(server): reuse WebSocket research log handler

RerankerGuo:
## Summary

Reuse the WebSocket command's initialized research log handler throughout the
streaming run.

## Problem

`handle_start_command()` creates a `CustomLogsHandler` and writes the query into
its JSON log. `WebSocketManager.start_streaming()` then calls `run_agent()`,
which created a second handler for the same task and file name.

The second constructor rewrote the JSON file with empty content, discarding the
query and any entrypoint metadata before research started.

## Changes

- Pass the existing handler from `handle_start_command()` to
  `WebSocketManager.start_streaming()`.
- Forward it to `run_agent()`.
- Reuse it when provided while retaining automatic handler creation for REST
  and other direct callers.
- Add propagation and no-second-construction regression tests.

## Verification

Before:

```text
3 failed, 5 passed
```

The entrypoint did not pass a handler, and the manager/runner did not accept it.

After:

```text
tests/test_websocket_manager.py .....  5 passed
tests/test_logging.py ...              3 passed
8 passed
```

The tests verify the full handler propagation chain and that the initialized
query remains in the JSON log.

local test process supplies those names without modifying that module.

## Impact

- **Breaking change:** No
- **WebSocket research logs:** Preserve entrypoint metadata
- **Direct/REST runs:** Continue creating their own handler when none is passed

## Checklist

- [x] The regression tests fail before and pass after the fix.
- [x] Existing WebSocket manager and logging tests pass.
- [x] The change is limited to log-handler lifecycle propagation.

### DIFF
--- a/backend/server/server_utils.py
+++ b/backend/server/server_utils.py
@@ -170,6 +170,7 @@ async def handle_start_command(websocket, data: str, manager):
         mcp_strategy,
         mcp_configs,
         max_search_results,
+        logs_handler=logs_handler,
     )
     report = str(report)
     file_paths = await generate_report_files(report, sanitized_filename)
--- a/backend/server/websocket_manager.py
+++ b/backend/server/websocket_manager.py
@@ -98,7 +98,7 @@ class WebSocketManager:
             except Exception:
                 pass  # If this fails too, there's nothing more we can do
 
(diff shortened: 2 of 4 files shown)


RerankerGuo:

PR #2031, opened 2026-08-04
fix(multi-agent): update md2pdf API usage

RerankerGuo:
## Summary

Keep multi-agent PDF publishing compatible with both the declared minimum and current `md2pdf` APIs.

## Problem

`multi_agents/agents/utils/file_formats.py` used the legacy `md_content` and `css_file_path` keywords. The locked `md2pdf 3.1.1` API renamed them to `raw` and `css`, so PDF export raised a `TypeError` that was caught internally and silently returned an empty path.

A direct replacement with only the 3.x keywords would break the project's declared minimum `md2pdf>=1.0.1`, whose verified signature still requires the legacy names.

## Changes

- Inspect the imported `md2pdf` callable and select its supported keyword set.
- Use `raw` and `css` for current releases.
- Preserve `md_content` and `css_file_path` compatibility for the declared minimum.
- Parameterize the regression test with strict legacy and current API stubs.

## Verification

Verified package source signatures:

```text
md2pdf 1.0.1: pdf_file_path, md_content, md_file_path, css_file_path, base_url
md2pdf 3.1.1: pdf, raw, md, css, base_url, extras, extras_config, context
```

Focused test:

```text
uv run --python 3.11 --with pytest --with pytest-asyncio \
  pytest tests/test_multi_agent_pdf_export_api.py -q
2 passed
```

The tests use strict function signatures, verify the generated output path and CSS, and do not import WeasyPrint.

## Impact

- **Breaking change:** No
- **Affected area:** Multi-agent PDF publishing
- **Supported dependency range:** `md2pdf>=1.0.1`, including locked 3.1.1
- **DOCX/Markdown publishing:** Unchanged

## Checklist

- [x] Current `md2pdf` parameters are covered.
- [x] Declared-minimum legacy parameters are covered.
- [x] No system PDF libraries are required by the regression tests.
- [x] The change remains limited to `md2pdf` API compatibility.

Re-audited against `main` at `5d84d2f`. `main` fails against the current strict API, while a current-only replacement fails against the declared minimum. This revision verifies and supports both API generations.

### DIFF
--- a/multi_agents/agents/utils/file_formats.py
+++ b/multi_agents/agents/utils/file_formats.py
@@ -57,8 +57,8 @@ async def write_md_to_pdf(text: str, path: str) -> str:
         # Moved imports to inner function to avoid known import errors with gobject-2.0
         from md2pdf.core import md2pdf
         md2pdf(file_path,
-               md_content=text,
-               css_file_path=css_path,
+               raw=text,
+               css=css_path,
                base_url=None)
         print(f"Report written to {file_path}")
     except Exception as e:
--- a/multi_agents/agents/utils/file_formats.py
+++ b/multi_agents/agents/utils/file_formats.py
@@ -1,8 +1,10 @@
-import aiofiles
+import inspect
+import os
 import urllib
 import uuid
+
+import aiofiles
 import mistune
-import os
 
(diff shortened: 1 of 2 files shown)


RerankerGuo:

PR #2037, opened 2026-08-04
fix(scraper): clean interrupted PDF downloads

RerankerGuo:
## Summary

Close streamed PDF responses and remove partial temporary files on every remote
PyMuPDF download path.

## Problem

`PyMuPDFScraper` requested remote PDFs with `stream=True` but never closed the
`requests.Response`. Its temporary-file cleanup also began only after
`iter_content()` finished.

If the connection failed after writing one or more chunks, the broad exception
handler returned the normal empty result while silently leaving both resources:

- the streamed HTTP response/socket remained open;
- the partial `.pdf` remained in the system temporary directory.

HTTP-status and parser failures also left the response unclosed.

## Changes

- Track the response and temporary path across the complete remote-download
  lifecycle.
- Close the response in `finally`.
- Remove a partially created PDF even when streaming is interrupted.
- Close an initial response before the existing SSL fallback retry.
- Extend the existing temporary-file regression suite with an interrupted
  stream case.

## Verification

Before:

```text
tests/test_pymupdf_tempfile_cleanup.py .F
assert response.closed
1 failed, 1 passed
```

After:

```text
tests/test_pymupdf_tempfile_cleanup.py .. 2 passed
```

Adjacent scraper tests:

```text
tests/test_pymupdf_tempfile_cleanup.py .. 2 passed
tests/test_scraper_get_scraper.py ......  6 passed
tests/test_arxiv_scraper.py ..            2 passed
10 passed
```

The tests use a process-local `Any`/`List` compatibility bootstrap for the
unrelated current-main import regression tracked by #1981; no source workaround
is included here.

Additional checks:

```text
uv run --with ruff ruff check \
  gpt_researcher/scraper/pymupdf/pymupdf.py \
  tests/test_pymupdf_tempfile_cleanup.py \
  --select F821
All checks passed

uv run --frozen python -m compileall -q <changed files>
```

## Duplicate Check

All 134 open PR titles, bodies, and changed files were checked immediately
before publishing. PRs #2002 and #1849 change the HTTP client selection so a
configured session/User-Agent is used. Neither closes responses nor cleans
partial files when `iter_content()` fails. This change does not alter client
selection and can be composed with either PR.

## Impact

- **Breaking change:** No
- **Affected area:** Remote PDF download resource lifecycle
- **Successful/local PDF behavior:** Unchanged
- **Network calls in tests:** None

## Checklist

- [x] The interrupted-stream regression fails before and passes after.
- [x] Response and temporary-file cleanup are both asserted.
- [x] Adjacent scraper tests pass.
- [x] The PR does not duplicate session/User-Agent changes.

### DIFF
--- a/gpt_researcher/scraper/pymupdf/pymupdf.py
+++ b/gpt_researcher/scraper/pymupdf/pymupdf.py
@@ -41,32 +41,37 @@ class PyMuPDFScraper:
         """
         try:
             if self.is_url():
+                response = None
+                temp_filename = None
                 try:
-                    response = requests.get(self.link, timeout=(5, 30), stream=True)
-                    response.raise_for_status()
-                except requests.exceptions.SSLError:
-                    import logging
-                    logging.getLogger(__name__).warning(
-                        f"SSL verification failed for {self.link}, retrying without verification"
-                    )
-                    response = requests.get(self.link, timeout=(5, 30), stream=True, verify=False)
-                    response.raise_for_status()
+                    try:
+                        response = requests.get(self.link, timeout=(5, 30), stream=True)
+                        response.raise_for_status()
+                    except requests.exceptions.SSLError:
+                        import logging
+                        if response is not None:
+                            response.close()
+                        logging.getLogger(__name__).warning(
+                            f"SSL verification failed for {self.link}, retrying without verification"
+                        )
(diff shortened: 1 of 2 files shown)


RerankerGuo:

PR #2041, opened 2026-08-05
fix(costs): schedule asynchronous cost log events

RerankerGuo:
## Summary

Deliver structured cost updates from the synchronous cost callback without
creating an unawaited coroutine.

## Problem

`GPTResearcher.add_costs()` called the asynchronous `_log_event()` helper
without awaiting or scheduling it. With structured logging enabled this caused:

```text
RuntimeWarning: coroutine 'GPTResearcher._log_event' was never awaited
```

The cost totals were updated, but `on_research_step("cost_update", ...)` was
never called.

## Changes

- Schedule cost log events on the active event loop.
- Retain background task references until each event completes.
- Preserve synchronous cost accumulation when no event loop is running.
- Add regression coverage for asynchronous and synchronous callers.

## Verification

Before:

```text
RuntimeWarning: coroutine 'GPTResearcher._log_event' was never awaited
Expected on_research_step to have been awaited once. Awaited 0 times.
```

After:

```text
uv run --python 3.11 python -m unittest tests.test_cost_logging -v
Ran 2 tests in 0.010s
OK

uvx ruff check gpt_researcher/agent.py tests/test_cost_logging.py \
  --select F821,RUF006
All checks passed!
```

The broader existing cost test module is currently blocked during collection
by the unrelated missing `Any`/`List` imports tracked in #1981 and PR #1988.

## Impact

- Breaking change: No
- Cost callback signature: unchanged
- Cost totals and per-step accounting: unchanged
- Structured cost events: now delivered

### DIFF
--- a/gpt_researcher/agent.py
+++ b/gpt_researcher/agent.py
@@ -164,6 +164,7 @@ class GPTResearcher:
         self.headers = headers or {}
         self.research_costs = 0.0
         self.step_costs: dict[str, float] = {}
+        self._background_tasks: set[asyncio.Task] = set()
         self._current_step: str = "general"
         self.log_handler = log_handler
         self.prompt_family = get_prompt_family(prompt_family or self.cfg.prompt_family, self.cfg)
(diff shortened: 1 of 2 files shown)


RerankerGuo:

Issue #1628, opened 2026-02-14
add support for gemini-3-flash-preview


valtahomes:
Hello,

Just tried to load the new gemini 3 model: gemini-3-flash-preview and seems like this is not supported yet. The content format is not a match.

Thanks,
V

assafelovic:
Model support like this is usually a small change — worth checking whether it works now, since the LLM provider layer has moved since February. If it still fails, the exact error would help: *"content format is not a match"* suggests a response-parsing mismatch rather than a missing model entry, and #2074 landed a batch of JSON-recovery fixes that may already cover it.

---

Context on timing: the PR queue was just drained from 129 to ~30 and the issue backlog reviewed, so this is being read rather than ignored. CI now runs on every PR — `imports` and `unit` across Python 3.11–3.14, plus `collect` — which makes a contribution here far easier to land than it was a week ago.

PR #2043, opened 2026-08-05
fix(llm): normalize structured content blocks

RerankerGuo:
## Summary

Normalize LangChain structured message content into plain text in both
streaming and non-streaming LLM paths.

Refs #1628.

## Problem

LangChain messages may expose `content` as either a string or a list of string
and dictionary blocks. The provider returned the list unchanged for
non-streaming calls and attempted to concatenate it to a string while
streaming:

```text
TypeError: can only concatenate str (not "list") to str
```

## Changes

- Normalize string and block-list content through one helper.
- Join string blocks and dictionary blocks containing text.
- Ignore non-text metadata/tool blocks.
- Apply the same behavior to streaming and non-streaming responses.

## Verification

Before:

```text
test_non_streaming_content_blocks_return_text ... FAIL
test_streaming_content_blocks_are_joined ... ERROR
TypeError: can only concatenate str (not "list") to str
```

After:

```text
uv run --python 3.11 python -m unittest tests.test_llm_content_blocks -v
Ran 2 tests
OK

uvx ruff check gpt_researcher/llm_provider/generic/base.py \
  tests/test_llm_content_blocks.py --select F821
All checks passed!
```

The adjacent usage-tracking test is currently blocked during collection by the
unrelated missing `Any`/`List` imports tracked in #1981 and PR #1988.

## Impact

- Breaking change: No
- Existing string responses: unchanged
- Structured text responses: returned as plain text
- Non-text content blocks: omitted from report text

### DIFF
--- a/gpt_researcher/llm_provider/generic/base.py
+++ b/gpt_researcher/llm_provider/generic/base.py
@@ -113,6 +113,24 @@ class ChatLogger:
                     "stacktrace": traceback.format_exc()
                 }) + "\n")
 
+
+def _content_to_text(content: Any) -> str:
+    """Normalize LangChain message content into plain text."""
+    if content is None:
+        return ""
+    if isinstance(content, str):
+        return content
+    if isinstance(content, list):
+        parts = []
+        for block in content:
+            if isinstance(block, str):
+                parts.append(block)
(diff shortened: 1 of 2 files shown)


RerankerGuo:

PR #2044, opened 2026-08-05
fix(document): offload synchronous parser work

RerankerGuo:
## Summary

Move synchronous LangChain document parsing off the asyncio event loop for
both local and downloaded documents.

## Problem

`DocumentLoader._load_document()` and
`OnlineDocumentLoader._load_document()` are asynchronous methods, but both
called the blocking `loader.load()` directly. PDF, Office, and HTML parsing can
therefore pause WebSocket handling and unrelated research tasks. The local
loader also uses `asyncio.gather()`, but the blocking calls prevented those
tasks from making concurrent progress.

## Changes

- Run local document parser calls with `asyncio.to_thread()`.
- Apply the same offload to downloaded document parsing.
- Preserve loader selection, errors, return values, and temporary-file cleanup.
- Add heartbeat regressions for both paths.

## Verification

Before:

```text
test_local_document_parsing_keeps_event_loop_responsive ... FAIL
test_online_document_parsing_keeps_event_loop_responsive ... FAIL
```

After:

```text
uv run --python 3.11 python -m unittest \
  tests.test_document_loader_offload -v
Ran 2 tests
OK

uvx ruff check gpt_researcher/document/document.py \
  gpt_researcher/document/online_document.py \
  tests/test_document_loader_offload.py --select F821,ASYNC251
All checks passed!
```

The existing URL-extension test is currently blocked during collection by the
unrelated missing `Any`/`List` imports tracked in #1981 and PR #1988.

## Duplicate Check

Open document PRs cover metadata fallback, error text, URL extension parsing,
EPUB support, or URL security. None offloads parser execution.

## Impact

- Breaking change: No
- Loader outputs and cleanup: unchanged
- Event loop responsiveness: preserved during parser work
- Parsing execution: moved to the default thread pool

### DIFF
--- a/gpt_researcher/document/document.py
+++ b/gpt_researcher/document/document.py
@@ -80,7 +80,7 @@ class DocumentLoader:
             loader = loader_dict.get(file_extension, None)
             if loader:
                 try:
-                    ret_data = loader.load()
+                    ret_data = await asyncio.to_thread(loader.load)
                 except Exception as e:
                     print(f"Failed to load HTML document : {file_path}")
                     print(e)
--- a/gpt_researcher/document/online_document.py
+++ b/gpt_researcher/document/online_document.py
@@ -1,3 +1,4 @@
+import asyncio
 import os
 import aiohttp
 import tempfile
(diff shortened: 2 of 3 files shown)


RerankerGuo:

PR #2045, opened 2026-08-05
fix(mcp): forward streaming log arguments

RerankerGuo:
## Summary

Forward MCP progress logs to the shared WebSocket streaming helper using its
actual argument names.

## Problem

`MCPStreamer.stream_log()` called `stream_output()` with `type`, `output`, and
`metadata` keyword arguments. The helper accepts `log_type`, `step`, `content`,
`with_data`, and `data`. Every WebSocket log therefore raised a `TypeError`
inside the catch block and was silently dropped:

```text
got an unexpected keyword argument 'type'
```

## Changes

- Map MCP log fields to the supported `stream_output()` signature.
- Preserve optional metadata via `with_data` and `data`.
- Add a strict-signature regression test for the complete payload.

## Verification

Before:

```text
Error streaming log: ... got an unexpected keyword argument 'type'
Expected one forwarded call; received zero.
```

After:

```text
uv run --python 3.11 python -m unittest tests.test_mcp_streaming -v
Ran 1 test
OK

uvx ruff check gpt_researcher/mcp/streaming.py \
  tests/test_mcp_streaming.py --select F821
All checks passed!
```

The test does not require an MCP server or network access.

## Impact

- Breaking change: No
- MCP execution: unchanged
- Server logs: unchanged
- WebSocket MCP progress events: delivered instead of dropped

### DIFF
--- a/gpt_researcher/mcp/streaming.py
+++ b/gpt_researcher/mcp/streaming.py
@@ -37,11 +37,12 @@ class MCPStreamer:
             try:
                 from ..actions.utils import stream_output
                 await stream_output(
-                    type="logs", 
-                    content="mcp_retriever", 
-                    output=message, 
+                    log_type="logs",
+                    step="mcp_retriever",
+                    content=message,
                     websocket=self.websocket,
-                    metadata=data
+                    with_data=data is not None,
+                    data=data,
                 )
             except Exception as e:
                 logger.error(f"Error streaming log: {e}")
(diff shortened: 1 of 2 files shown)


RerankerGuo:
Closing after validating the integration against the real imported helper. `MCPStreamer` imports `gpt_researcher.actions.utils.stream_output`, whose signature is `type, content, output, websocket, output_log, metadata`; the current `main` call already matches it and forwards the expected WebSocket payload. This PR changed the call to the incompatible signature from `gpt_researcher.retrievers.utils.stream_output`, causing `unexpected keyword argument log_type` and dropping the payload. The regression test mocked that incompatible signature, so it did not exercise the actual integration. No upstream change is needed.

PR #2046, opened 2026-08-05
fix(server): offload report store file I/O

RerankerGuo:
## Summary

Move ReportStore JSON file operations off the asyncio event loop.

## Problem

The ReportStore API is asynchronous and is called directly by FastAPI routes,
but it performed synchronous `Path.read_text()`, JSON parsing/serialization,
`Path.write_text()`, and atomic replacement while holding its asyncio lock.
Large report histories or slow filesystems could therefore pause WebSocket and
HTTP handling for the entire event loop.

## Changes

- Offload the complete read and JSON parsing operation with
  `asyncio.to_thread()`.
- Offload directory creation, serialization, temporary write, and atomic
  replacement as one operation.
- Preserve the existing asyncio lock and atomic replacement behavior.
- Add heartbeat regressions for read and write paths.

## Verification

Before:

```text
test_read_does_not_block_event_loop ... FAIL
test_write_does_not_block_event_loop ... FAIL
```

After:

```text
uv run --python 3.11 python -m unittest \
  tests.test_report_store_async_io -v
Ran 2 tests
OK

uvx ruff check backend/server/report_store.py \
  tests/test_report_store_async_io.py --select F821,ASYNC251
All checks passed!
```

The tests use real temporary JSON files with delayed `Path` operations and do
not require the server or network.

## Impact

- Breaking change: No
- Report data format: unchanged
- Locking and atomic writes: preserved
- Event loop responsiveness: preserved during report store I/O

### DIFF
--- a/backend/server/report_store.py
+++ b/backend/server/report_store.py
@@ -9,25 +9,31 @@ class ReportStore:
         self._path = path
         self._lock = asyncio.Lock()
 
-    async def _ensure_parent_dir(self) -> None:
-        self._path.parent.mkdir(parents=True, exist_ok=True)
-
     async def _read_all_unlocked(self) -> Dict[str, Dict[str, Any]]:
-        if not self._path.exists():
+        def read_all() -> Dict[str, Dict[str, Any]]:
+            if not self._path.exists():
+                return {}
+            try:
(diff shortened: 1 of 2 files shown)


RerankerGuo:

PR #2048, opened 2026-08-05
feat: add Sciverse retriever for scientific literature full-text search

gary-Shen:
Sciverse indexes paper metadata and full text, so a search returns passages from paper bodies rather than abstracts alone. Results map to the standard title/href/body contract, with href preferring the open-access URL and falling back to the DOI so citations stay resolvable.

Registered in retrievers/__init__.py, VALID_RETRIEVERS and the get_retriever factory. Uses requests, matching the other academic retrievers, so no new dependencies are required.

### DIFF
--- a/docs/docs/gpt-researcher/search-engines/search-engines.md
+++ b/docs/docs/gpt-researcher/search-engines/search-engines.md
@@ -46,6 +46,7 @@ Thanks to our community, we have integrated the following web search engines and
 - [Exa](https://docs.exa.ai/reference/getting-started) - Env: `RETRIEVER=exa`
 - [fastCRW](https://fastcrw.com/docs/rest-api) - Env: `RETRIEVER=crw`
 - [PubMedCentral](https://www.ncbi.nlm.nih.gov/home/develop/api/) - Env: `RETRIEVER=pubmed_central`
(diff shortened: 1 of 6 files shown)


assafelovic:
Reviewed properly — thanks for a well-formed PR.

It applies cleanly to current `main` and the full suite passes with it (400 tests, 0 failures — your own tests included). Error handling is there, and it follows the existing retriever shape.

One gap: no docs entry. The other retriever PRs update `docs/docs/gpt-researcher/search-engines/search-engines.md`; this one doesn't, so `sciverse` would be undiscoverable.

---

gary-Shen:
Thanks for the thorough review — and for stating the real blocker plainly instead of letting it sit.

Docs gap is fixed: sciverse is now listed in search-engines.md with its env vars.

Also adapted to the requires_scraping contract that landed in [#2079](https://github.com/assafelovic/gpt-researcher/pull/2079) — sciverse declares False and passes the new contract suite.

On scope: the entry-point plugin path sounds right to us, and we'd rather see that land than have this merged as the 23rd retriever against the project's better judgment. If @assafelovic goes that way, count us in as a first adopter — we'd ship and version a gpt-researcher-sciverse package ourselves, which frankly suits a vendor better than parking the maintenance on your side: our API, our response shapes, our release cadence, and your core stays at 22.

Two notes for whichever way it goes:

The tests are fully mocked and key-free (CI never touches our API), and the fixtures mirror the exact response shapes we verified against the live service — including the second link-resolution call — so this shouldn't add to the "unexpected payload shape" pile.
Either way, we maintain this integration: if it merges and our API ever changes shape, the fix PR comes from us.

Happy to keep this open as a test case for the plugin discussion, or convert it to the first external package once an entry-point group exists.

PR #2054, opened 2026-08-07
Update embeddings.py

MaxAkaAltmer:
Fixed selection of local embedding model.


assafelovic:

MaxAkaAltmer:
Issue with embedding_kwargs["openai_api_base"] also fixed? Or I need to use some workaround for using embedder?

Issue #2055, opened 2026-08-07
v3.6.0 broken as shipped: stray committed gitlink breaks git-URL Docker builds; missing typing import crashes the server at boot


NyKalC:
Thanks for the project — we run it self-hosted (fully local: Ollama LLMs + SearXNG retriever) and hit two independent defects in the `v3.6.0` release (commit `5d84d2f5553e70a2765a8ff3a0d2672d60437ce8`). Both look like artifacts of a release cut from a dirty working tree. Details and one-line fixes below.

### Defect 1 — stray committed gitlink breaks Docker git-URL builds

The v3.6.0 tree contains a gitlink (submodule-mode entry) with no corresponding `.gitmodules`:

```
$ git ls-files -s | awk '$1==160000'
160000 6c75117d27b866163aac44ce0e2360958ffbe825 0   .claude/worktrees/crazy-curie-e43899

$ ls .gitmodules
ls: cannot access '.gitmodules': No such file or directory
```

Docker/BuildKit git-URL build contexts initialize submodules unconditionally, so any build that consumes the repo as a git URL hard-fails:

```
$ docker build https://github.com/assafelovic/gpt-researcher.git#v3.6.0
...
fatal: No url found for submodule path '.claude/worktrees/crazy-curie-e43899' in .gitmodules
```

The same applies to any CI/workflow that runs `git submodule update --init`. (Plain `git clone` still works — the gitlink is just left uninitialized as an empty directory.)

**Fix:** `git rm --cached .claude/worktrees/crazy-curie-e43899` and commit; consider adding `.claude/` to `.gitignore` so a local agent worktree can't be committed again.

### Defect 2 — `query_processing.py` missing its typing import; the server cannot boot

`gpt_researcher/actions/query_processing.py` in v3.6.0 uses `Any` / `Dict` / `List` in annotations but never imports them (and has no `from __future__ import annotations`), e.g.:

```python
def _normalize_sub_queries(parsed: Any, fallback_query: str) -> List[str]:
```

Annotations are evaluated at import time, so the backend dies at startup:

```
NameError: name 'Any' is not defined
```

Minimal repro without Docker:

```
pip install git+https://github.com/assafelovic/gpt-researcher.git@v3.6.0
python -c "import gpt_researcher.actions.query_processing"
```

[...]

PR #2063, opened 2026-08-12
fix(multi_agents): bind _route_fact_check and _route_draft_review to their free functions

### LINKED ISSUE #2060: multi_agents: both review-loop routers reference undefined methods — workflow crashes at graph construction (regression around #1909)
## Summary

At current master (checked at `5d84d2f`), the `multi_agents` workflow cannot be constructed: both conditional-edge routers are wired to methods that do not exist.

1. **Fact-check loop.** `multi_agents/agents/orchestrator.py`, line 102:

   ```python
   workflow.add_conditional_edges(
       'fact_checker',
       self._route_fact_check,
       {"accept": "visualizer", "revise": "writer"}
   )
   ```

   `ChiefEditorAgent` defines no `_route_fact_check` (no base class, no assignment anywhere in the package) — `self._route_fact_check` raises `AttributeError` the moment `_add_workflow_edges` runs, i.e. on every `run_research_task` call.

2. **Draft-review loop.** `multi_agents/agents/editor.py`, line 140: the inner reviewer/reviser workflow wires `self._route_draft_review`; `EditorAgent` defines no such method either.

The intended routers appear to exist already: `plan_review.route_human_feedback`, `fact_review.route_fact_check`, and `draft_review.route_draft_review` are complete, documented free functions (with bounds and graceful-exceed errors) — but nothing imports `fact_review` or `draft_review`, and the two `self._route_*` bindings that would call them are missing. It looks like the modules landed in the #1909 era while the one-line bindings did not.

## Reproduction sketch

```python
from multi_agents.agents import ChiefEditorAgent
ChiefEditorAgent(task={"query": "x"}).init_research_team()
# AttributeError: 'ChiefEditorAgent' object has no attribute '_route_fact_check'
```

(And after fixing that one, the same again for `_route_draft_review` when the editor's inner workflow is built.)

## Suggested fix

Bind the existing routers where they were meant to be used:

```python
# orchestrator.py
from .fact_review import route_fact_check
def _route_fact_check(self, state):
    return route_fact_check(state, self.task.get("max_fact_check_revisions"))

# editor.py
from .draft_review import route_draft_review
def _route_draft_review(self, draft):
    return route_draft_review(draft, ...)
```

A one-line smoke test that merely constructs the workflow (`init_research_team()` + `compile()`) in CI would keep this class of regression from shipping.

AmirF194:
## Root cause

`ChiefEditorAgent._add_workflow_edges` wires the fact-checker loop to `self._route_fact_check`, and `EditorAgent._create_workflow` wires the reviewer loop to `self._route_draft_review`. Neither method is defined anywhere in the class. Building either agent's graph raises `AttributeError` immediately, before a single node runs:

```
AttributeError: 'ChiefEditorAgent' object has no attribute '_route_fact_check'
```

Both loops already have a fully implemented, unit-tested free function (`route_fact_check` in `fact_review.py`, `route_draft_review` in `draft_review.py`); they were never bound to the class. Git history on both fault lines shows a working inline lambda was replaced by the unbound `self._route_*` call in a same-day cherry-pick (`3dcb8392`, `5b59b615`), and the method definition never landed.

## Fix

Add `_route_fact_check` on `ChiefEditorAgent` and `_route_draft_review` on `EditorAgent`, mirroring the existing `_route_human_feedback` pattern: read the revision ceiling from the task dict (defaulting to the free function's own constant) and delegate.

## Verification

- New test `tests/test_multi_agents_route_bindings.py`: 6 cases covering both methods (bound + delegates correctly, respects a task-level revision-count override, and the exact `_create_workflow`/`_add_workflow_edges` repro from this issue). All 6 fail with the `AttributeError` above on `main`, all 6 pass on this branch.
- `tests/test_multi_agents_fact_revisions.py`, `test_multi_agents_draft_revisions.py`, `test_multi_agents_plan_revisions.py` still pass unmodified.
- Not checked: the repo currently fails to `import gpt_researcher` at all on this HEAD, independently of this change, due to an unordered `typing` import (#2055); the new test avoids that path entirely by loading `orchestrator.py`/`editor.py` directly rather than through the package.

Fixes #2060

### DIFF
--- a/multi_agents/agents/editor.py
+++ b/multi_agents/agents/editor.py
@@ -8,6 +8,10 @@ from .utils.views import print_agent_output
 from .utils.llms import call_model
 from ..memory.draft import DraftState
 from . import ResearchAgent, ReviewerAgent, ReviserAgent
+from .draft_review import (
+    DEFAULT_MAX_DRAFT_REVISIONS,
+    route_draft_review,
+)
 
 
 class EditorAgent:
@@ -142,6 +146,12 @@ class EditorAgent:
         )
 
         return workflow
+
+    def _route_draft_review(self, draft: Dict[str, any]) -> str:
+        task = draft.get("task") or {}
+        max_draft_revisions = task.get(
+            "max_draft_revisions", DEFAULT_MAX_DRAFT_REVISIONS)
+        return route_draft_review(draft, max_draft_revisions)
 
     def _log_parallel_research(self, queries: List[str]) -> None:
         """Log the start of parallel research tasks."""
--- a/multi_agents/agents/orchestrator.py
+++ b/multi_agents/agents/orchestrator.py
@@ -10,6 +10,10 @@ from .plan_review import (
     DEFAULT_MAX_PLAN_REVISIONS,
     route_human_feedback,
 )
+from .fact_review import (
+    DEFAULT_MAX_FACT_CHECK_REVISIONS,
+    route_fact_check,
+)
 
 # Import agent classes
 from . import \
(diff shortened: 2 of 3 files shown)


AmirF194:
No rush, just flagging this is still open. Mergeable and no maintainer has looked yet; the one CI check is held pending approval for a first-time contributor, so it hasn't run at all.

assafelovic:

AmirF194:
Thanks for taking the time to fold it in and for the credit. Glad it's part of a healthy main again.

PR #2066, opened 2026-08-14
fix(readme): restore broken star history chart

Dessalines39394:
The star history chart shown in the README and its localized translations is currently broken and renders no image, since the upstream service stopped working because of GitHub stargazer API restrictions.

This change points the chart to a maintained mirror that serves the same chart reliably, restoring the visualization across the main README and the Japanese, Korean and Chinese translations. No functional or documented behavior changes.

### DIFF
--- a/README-ja_JP.md
+++ b/README-ja_JP.md
@@ -169,11 +169,11 @@ GPT Researcher は **LangSmith** をサポートしており、複雑なマル�
 ---
 
 <p align="center">
-<a href="https://star-history.com/#assafelovic/gpt-researcher">
+<a href="https://star-history.dera.page/#assafelovic/gpt-researcher">
   <picture>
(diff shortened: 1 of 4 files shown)


assafelovic:
Verified and merging — thanks.

I initially checked this the lazy way (HTTP status on `api.star-history.com`, which returns 200) and nearly dismissed it. Reading the actual SVG body shows you're right: it renders the text *"GitHub restricted access to star data"* instead of a chart, per their [blog post](https://star-history.com/blog/github-stargazer-api-restriction). The replacement returns a real chart — axis labels 5k–25k, 2024–2026, ten plotted paths.

One thing worth flagging for the record rather than blocking on: this points the README images at `dera.page`, which isn't the well-known star-history.com domain, so it's a third-party host serving an image on the project's front page in four languages. Given the original is broken upstream and this works, it's a fair trade — but worth revisiting if star-history.com restores the API.

PR #2068, opened 2026-08-18
Add Serply web and Google Scholar retrievers

googio:
## Add Serply retrievers (web + Google Scholar)

This adds two optional retrievers backed by the [Serply](https://serply.io/) SERP API, following the existing per-provider pattern (one directory under `gpt_researcher/retrievers/`, registered in `retrievers/__init__.py`, `retrievers/utils.py`, and `actions/retriever.py`):

- `RETRIEVER=serply` - Google web results, mirroring the `serper` retriever it sits next to.
- `RETRIEVER=serply_scholar` - Google Scholar results. GPT Researcher already has arXiv, Semantic Scholar, PubMed Central, and OpenAlex, but no Google Scholar, which spans publishers, theses, and citations across every field. This gives the research flow that coverage.

Both read `SERPLY_API_KEY` from the environment and normalize responses to the standard `{title, href, body}` shape. Like the sibling retrievers, `search()` always returns a list, never `None`.

### Why two

The repo already pairs same-vendor SERP wrappers (`serper` and `serpapi`) and ships four academic retrievers, so both variants fit the existing surface. Each is small and self-contained; the file count comes from the per-provider scaffold applied twice, not from a large change to any one file.

### Optional and non-breaking

Nothing changes for anyone who does not set `RETRIEVER=serply` or `serply_scholar`. No new dependency is added (`requests` is already required). Note: Serply sits behind Cloudflare, which rejects the default `requests` User-Agent, so both retrievers send an explicit one.

### Testing

- Live-tested both retrievers against `api.serply.io`, confirming real web and scholar results normalize to the expected shape.
- Added `tests/test_serply_returns_list.py`, mirroring `test_serper_returns_list.py`, covering the return-a-list error paths with the network mocked.

Docs: added both entries to `docs/docs/gpt-researcher/search-engines/search-engines.md`. More at [serply.io/docs](https://serply.io/docs).

Disclosure: I work with Serply. Happy to adjust scope, naming, or split the web and scholar retrievers into separate PRs if you prefer.

### DIFF
--- a/docs/docs/gpt-researcher/search-engines/search-engines.md
+++ b/docs/docs/gpt-researcher/search-engines/search-engines.md
@@ -30,6 +30,8 @@ Thanks to our community, we have integrated the following web search engines:
 - [SearchApi](https://www.searchapi.io/) - Env: `RETRIEVER=searchapi`
 - [Serp API](https://serpapi.com/) - Env: `RETRIEVER=serpapi`
 - [Serper](https://serper.dev/) - Env: `RETRIEVER=serper` - [Setup Guide](#serper)
+- [Serply](https://serply.io/) - Env: `RETRIEVER=serply` and `SERPLY_API_KEY` - [Docs](https://serply.io/docs)
+- [Serply Scholar](https://serply.io/) - Google Scholar results - Env: `RETRIEVER=serply_scholar` and `SERPLY_API_KEY` - [Docs](https://serply.io/docs)
 - [Searx](https://searx.github.io/searx/) - Env: `RETRIEVER=searx`
 - [Duckduckgo](https://pypi.org/project/duckduckgo-search/) - Env: `RETRIEVER=duckduckgo`
 - [Arxiv](https://info.arxiv.org/help/api/index.html) - Env: `RETRIEVER=arxiv`
--- a/gpt_researcher/actions/retriever.py
+++ b/gpt_researcher/actions/retriever.py
@@ -20,6 +20,8 @@ def get_retriever(retriever: str):
         - searchapi: SearchAPI service
         - serpapi: SerpAPI service
         - serper: Serper API
+        - serply: Serply SERP API (Google web results)
+        - serply_scholar: Serply SERP API (Google Scholar results)
         - duckduckgo: DuckDuckGo search
         - bing: Bing search
         - brave: Brave Search API
@@ -56,6 +58,14 @@ def get_retriever(retriever: str):
             from gpt_researcher.retrievers import SerperSearch
 
             return SerperSearch
+        case "serply":
+            from gpt_researcher.retrievers import SerplySearch
+
+            return SerplySearch
+        case "serply_scholar":
+            from gpt_researcher.retrievers import SerplyScholarSearch
+
+            return SerplyScholarSearch
         case "duckduckgo":
             from gpt_researcher.retrievers import Duckduckgo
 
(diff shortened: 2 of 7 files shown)


assafelovic:
Reviewed properly — thanks, this is a clean PR.

It applies to current `main` and the full suite passes with it (392 tests, 0 failures). Tests, docs and error handling are all present, and adding both the web and Scholar variants in one PR is the right granularity.

---

googio:
That is a fair call, and the entry-point path is the right shape. I would like to build it, if you want it built.

The obvious caveat first: I am the vendor on this PR, so I am proposing work that would make this PR unnecessary. That is fine by me. The mechanism below has no Serply in it, I am not asking for #2068 in exchange, and it stands whether this one merges, sits, or closes.

## What the seam looks like today

Three places decide whether a retriever name is real, and they disagree about how they know:

- `actions/retriever.py` `get_retriever()` is a `match` with one `case` per name, ending in `case _: return None`.
- `config/config.py:194` `parse_retrievers()` validates against `get_all_retriever_names()` and raises `ValueError` on an unknown name, so it rejects first and dispatch is never reached.
- `retrievers/utils.py` `get_all_retriever_names()` is an `os.listdir` of the `retrievers/` package directory, falling back to the hardcoded `VALID_RETRIEVERS` in its except branch.

The directory scan is why "ship it outside the tree" does not work today: an installed package is not a subdirectory of `gpt_researcher/retrievers/`, so it fails the config gate before anything gets a chance to import it.

## What I would add

An entry-point group, say `gpt_researcher.retrievers`, read through `importlib.metadata`:

- `get_retriever()` keeps the `match` byte-identical and consults entry points only in the `case _` branch. Every in-tree name resolves exactly as it does now and cannot be shadowed by an installed package.
- `get_all_retriever_names()` returns the directory scan unioned with the installed entry-point names, so `parse_retrievers()` stops rejecting them.
- A vendor ships `[project.entry-points."gpt_researcher.retrievers"]` with `foo = "gpt_researcher_foo:FooSearch"` and owns their own release cadence, their own key, and their own breakage.

No new dependency: `importlib-metadata>=8.7.0` is already in `pyproject.toml` and `requires-python` is `>=3.11`.

## The part that actually matters

You named payload shapes as the recurring cost, and I think that is the real work here rather than the lookup, which is twenty lines.

`docs/docs/gpt-researcher/search-engines/test-your-retriever.md` already states the contract in prose, "title, body, and href". The tree does not quite hold to it:

- `href` and `body` are the two keys everything agrees on. `tavily`, `exa` and `duckduckgo` emit exactly those two, and the duckduckgo module carries a comment about normalizing the newer `ddgs` shape back to them.
- `title` is in the doc and `serper` emits it, but `tavily` and `exa` do not.
- `CustomRetriever` returns `[{"url": ..., "raw_content": ...}]`, a different contract in the same tree, undetected.

That drift is absorbable while every retriever is yours. It stops being absorbable the moment third parties ship their own packages against an unwritten shape, which is the failure mode you already paid for twenty times.

So I would write the protocol down as part of this and give it a machine check: `__init__(query, query_domains=None, **kwargs)` and `search(max_results) -> list[dict]` with the required keys settled, as a conformance helper a vendor imports and runs against their own class, plus one test running it over the in-tree retrievers so the tree cannot drift from its own doc again. Not a runtime validator on the hot path unless you want one.

Whether `title` is required or optional, and whether `CustomRetriever` gets normalized or the contract widens to accept both, are your calls and I would rather ask than pick.

## Scope questions that are yours, not mine

1. Does the group resolve a class, or a zero-argument factory returning one?
2. Does an entry point whose import raises get logged and skipped, or propagate?
3. Do any in-tree retrievers move out later, or is this additive only? I would not touch a single existing one in this PR either way.

I have not written any of it and will not until you say go. If you would rather build it yourself, or decide the plugin path is not where you want this to land, that is a real answer and I will not chase it.

PR #2070, opened 2026-08-19
fix: apply OpenAI prompt-cache discount to cache_read tokens (#2065)

### LINKED ISSUE #2065: OpenAI cost tracking never reads input_token_details.cache_read, so prompt-cache-hit tokens are billed at full price
**Describe the bug**

`_extract_usage_tokens` in `gpt_researcher/utils/costs.py` only reads the flat `input_tokens`/`output_tokens` fields off LangChain's standardized `usage_metadata`, and `calculate_llm_cost`'s OpenAI branch prices the full `input_tokens` count at the standard rate. Neither ever looks at `input_token_details.cache_read`, the portion of input tokens that hit OpenAI's automatic prompt caching and are billed at a discount. Any research run that benefits from provider-side caching (gpt-researcher's large repeated system prompts routinely trigger it, since caching kicks in on any prompt over roughly 1024 tokens repeated within the cache window) gets its cost overreported.

This is the same class of gap as #1986/#1989 (Anthropic cache tokens priced at $0), but on the OpenAI/generic branch, which #1989's diff does not touch, and in the opposite direction: overcounted here instead of undercounted there.

**To Reproduce**

```python
import importlib.util
spec = importlib.util.spec_from_file_location("costs", "gpt_researcher/utils/costs.py")
costs = importlib.util.module_from_spec(spec)
spec.loader.exec_module(costs)

usage_no_detail = {"input_tokens": 5000, "output_tokens": 200}
usage_with_cache = {
    "input_tokens": 5000,
    "output_tokens": 200,
    "input_token_details": {"cache_read": 4500},
}

print(costs.calculate_llm_cost("openai", "gpt-4o", "", "", usage_metadata=usage_no_detail))
print(costs.calculate_llm_cost("openai", "gpt-4o", "", "", usage_metadata=usage_with_cache))
```

Both calls print `0.0145`, even though the second usage dict reports 4500 of the 5000 input tokens as cache hits. Ran this in a clean `python:3.11-slim` container (`pip install tiktoken` only, no other deps) against current `master` (`5d84d2f`).

**Expected behavior**

The cached portion of `input_tokens` should be priced at OpenAI's discounted cache rate, the way `#1989` now does for Anthropic's `cache_read_input_tokens`.

**Additional context**

I have a PR open against this repo already and would rather not stack a second one on the same file right now, so filing this as a report. Happy to send a PR for it if useful.

aranya-chatterjee:
## Problem

`_extract_usage_tokens` in `gpt_researcher/utils/costs.py` only read the
flat `input_tokens`/`output_tokens` fields off LangChain's standardized
`usage_metadata`, and `calculate_llm_cost`'s OpenAI branch priced the
full `input_tokens` count at the standard rate. Neither looked at
`input_token_details.cache_read`, the portion of input tokens that hit
OpenAI's automatic prompt caching and are billed at a discount.

Any research run that benefits from provider-side caching gets its cost
overreported.

This is the same class of gap as #1986/#1989 (Anthropic cache tokens
priced at $0), but on the OpenAI/generic branch, in the opposite
direction: overcounted here instead of undercounted there.

## Fix

- `_extract_usage_tokens` now also extracts `input_token_details.cache_read`
  and returns it alongside input/output tokens.
- `calculate_llm_cost` splits `input_tokens` into its cached and
  non-cached portions, pricing the cached portion at
  `OPENAI_CACHED_INPUT_DISCOUNT` (50%, per OpenAI's pricing page) instead
  of the full input rate.
- Applies to both the known-model pricing table and the flat-rate
  fallback path.

## Testing

Added `TestOpenAICachedInputPricing` to `tests/test_costs.py`, covering:
- No cache details (unaffected, matches pre-fix behavior)
- Partial cache hit (reproduces the exact scenario from #2065)
- Explicit `cache_read: 0` (matches baseline)
- Fully cached input
- Unknown-model fallback (cache discount still applies)

Verified against the reporter's exact repro from #2065:
- Before: both calls returned `0.0145` regardless of cache hits
- After: `0.0145` (no cache) vs `0.008875` (4500/5000 tokens cached)

Fixes #2065

### DIFF
--- a/gpt_researcher/utils/costs.py
+++ b/gpt_researcher/utils/costs.py
@@ -15,6 +15,14 @@ OUTPUT_COST_PER_TOKEN = 0.000015
 IMAGE_INFERENCE_COST = 0.003825
 EMBEDDING_COST = 0.02 / 1000000  # Assumes new ada-3-small
 
+# OpenAI's automatic prompt caching bills the cached portion of input
+# tokens at a discount off the standard input rate (currently 50% across
+# the GPT-4o/4.1/5 families). Kept as a single constant since OpenAI has
+# used a flat 50% cache discount across all its cached models to date;
+# revisit if a future model introduces a different ratio.
+# https://openai.com/api/pricing/
+OPENAI_CACHED_INPUT_DISCOUNT = 0.5
+
 logger = logging.getLogger(__name__)
 
 # (patterns, input $/MTok, output $/MTok) - first match wins, so more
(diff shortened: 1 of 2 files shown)


assafelovic:

FlorianFranzen:
@aranya-chatterjee Please be aware that despite what the LLM generated comment wants you to believe, your change has not been attributed with the appropriate `Co-Authored-By:` byline when it was merged upstream.

PR #2073, opened 2026-08-22
fix: import-time NameError in query_processing and missing tavily-python dependency

srj-17:
## Summary
Two small fixes for errors arising while getting the project running from a fresh install:

### 1. Import-time crash in `gpt_researcher/actions/query_processing.py`

`_normalize_sub_queries()` annotates parameters with `Any` / `List[str]`,
but the module's only `typing` import sat *below* the function definition.
Since annotations are evaluated at `def` time, importing anything that
pulls in this module raised:

 NameError: name 'Any' is not defined

**Fix:** consolidated into a single `from typing import Any, Dict, List`
at the top of the file. Also drops a redundant `f` prefix on a
`logger.warning` string with no placeholders (fixed by linter).

### 2. Missing default-retriever dependency in `pyproject.toml`

`tavily-python` was only declared in `requirements.txt`, so 
installing using `uv` missed it entirely.

**Fix:** added `"tavily-python>=0.7.12"` - pin matches `requirements.txt`

## Impact evidence

Full-suite pytest on current `main`: **41 collection errors** (`NameError:
name 'Any' is not defined`, propagated through every module importing
`query_processing`). 
On this branch: **1 error** — a pre-existing, unrelated
ImportError in `tests/test_security_fix.py` (missing `secure_filename`,
present on unmodified `main` as well).

### DIFF
--- a/gpt_researcher/actions/query_processing.py
+++ b/gpt_researcher/actions/query_processing.py
@@ -1,6 +1,7 @@
 import json_repair
 
 from gpt_researcher.llm_provider.generic.base import ReasoningEfforts
+from typing import Any, Dict, List
 
 
 def _normalize_sub_queries(parsed: Any, fallback_query: str) -> List[str]:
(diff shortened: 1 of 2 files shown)


assafelovic:

Issue #2069, opened 2026-08-18
Import should be at the top of the file right?


MelvinLoos:
I am not a python developer so I might be mistaken but this was throwing an error when booting the project through docker compose.
Shouldn't all imports be at the top of the file? Especially when the import for Any is defined underneath the place it is actually being used/called upon. If I remembered correctly Python is executed from top to bottom so order would matter and moving it to the top fixed the error so I guess so. 

https://github.com/assafelovic/gpt-researcher/blob/5d84d2f5553e70a2765a8ff3a0d2672d60437ce8/gpt_researcher/actions/query_processing.py#L35-L39

Looking at the commit history this just seems to be some merging mistake but I am surprised there are no CI jobs running to fix this very obvious bug. Just starting the app, not even running tests, would catch this immediately. 

Want me to send in a PR for the imports? (not CI)

assafelovic:
Fixed in `main` via #2074 (commit `bea0ad07`, from @MrSampson's #1943).

You were right, and it was worse than a style issue: `_normalize_sub_queries` was annotated with `Any`/`List` at line 6 while `typing` was imported at line 37. Annotations evaluate eagerly on Python ≤3.13, so the module raised `NameError` at import — taking down the library, the server and the published wheel. It only looked fine on 3.14, where PEP 649 defers annotation evaluation.

The imports are now at the top, there's a regression test asserting the package imports, and CI runs an `imports` job across 3.12/3.13/3.14 so this cannot silently come back.

Issue #2062, opened 2026-08-11
multi_agents: plan-revision limit can never trigger — router reads `revisions_count`, which nothing writes (and two competing bounds disagree)


alleytom1:
## Summary

The human plan-review loop's escape hatch is dead code, so the loop is effectively unbounded (until LangGraph's recursion limit kills the run).

1. **The router reads a key nothing writes.** `multi_agents/agents/orchestrator.py`, lines 88–96:

   ```python
   MAX_REVISIONS = 5
   workflow.add_conditional_edges(
       'human',
       lambda state: (
           "accept" if state['human_feedback'] is None
           else "force_accept" if state.get('revisions_count', 0) >= MAX_REVISIONS
           else "revise"
       ),
       {"accept": "researcher", "force_accept": "researcher", "revise": "planner"}
   )
   ```

   The state schema (`memory/research.py`) and `human.py` maintain `plan_revision_count`; no code in the package ever writes `revisions_count`. `state.get('revisions_count', 0)` is therefore always `0`, `force_accept` can never fire, and a user who keeps giving feedback loops planner→human until `GraphRecursionError`.

2. **Two competing implementations, neither effective.** `plan_review.route_human_feedback` (added in the #1909 era) reads the *correct* key and enforces `DEFAULT_MAX_PLAN_REVISIONS = 3` with a graceful error — and `ChiefEditorAgent._route_human_feedback` correctly delegates to it — but the conditional edge above uses the inline lambda instead, with a different constant (5 vs 3). Two sources of truth for the same bound; the wired one is broken, the correct one is unreachable.

## Reproduction sketch

Run with `include_human_feedback: true` and reply with substantive feedback more than 5 times: `force_accept` never triggers (counter stuck at 0); the run dies at the recursion limit instead of gracefully force-accepting.

## Suggested fix

Replace the inline lambda with the existing method (one line):

```python
workflow.add_conditional_edges(
    'human',
    lambda state: self._route_human_feedback(state),  # -> plan_review.route_human_feedback
    {"accept": "researcher", "force_accept": "researcher", "revise": "planner"}
)
```

…after aligning `route_human_feedback`'s return values with the edge map (it currently returns accept/revise and raises on exceed; either map the exception to `force_accept` or return `force_accept` at the limit). That also collapses the 3-vs-5 divergence to a single configurable `max_plan_revisions`.

Vedant-Git-dev:
Hi, I'd like to take this on.

[...]

PR #2051, opened 2026-08-06
Fix #2047: Add jsdom override to resolve Docker/ESM build failure

itsnevu:
Fixes #2047

Adds an `overrides` entry in `frontend/nextjs/package.json` to pin `jsdom` to `25.0.1`, which resolves the ESM/CommonJS incompatibility causing the frontend build to fail inside Docker as described in the issue.

PR #2074, opened 2026-08-23
fix: restore main, add CI, and land 97 queued PRs

assafelovic:
Restores `main`, then drains most of the PR queue behind it.

`main` currently ships two crash regressions, both introduced by the 2026-07-14 consolidation, and neither is caught by CI because **no CI job has ever run pytest**.

## What was broken

**The published package crashes on import for every supported Python except 3.14.** `gpt_researcher/actions/query_processing.py` annotated `_normalize_sub_queries(parsed: Any, ...) -> List[str]` at line 6 but imported `typing` at line 37. Annotations evaluate eagerly on ≤3.13, so the module raised `NameError` at import — taking down the library, the server and the PyPI wheel. It was invisible on 3.14, where PEP 649 defers annotation evaluation. Everything we ship is not: `Dockerfile` pins 3.12, `Dockerfile.fullstack` 3.13.3, `backend/runtime.txt` 3.12. Reported in #1981, #2055, #2069; **ten separate PRs** were opened against this one file.

**The multi-agent flow crashed before it started.** `orchestrator.py` wired `self._route_fact_check` and `editor.py` wired `self._route_draft_review`; neither method existed, so `init_research_team()` raised `AttributeError` and `multi_agents` was dead on arrival. Reported in #2060.

**A stray `160000` gitlink** (`.claude/worktrees/crazy-curie-e43899`) pointed at an unreachable commit, breaking recursive clones and git-URL Docker builds. Reported in #2055.

## Found while auditing — no open PR covered these

- `Procfile` launched `backend.server.server:app`, a module that does not exist. Procfile deploys failed at boot.
- `backend/server/server_utils.py` did a bare `from utils import ...` that resolved only because `app.py` mutates `sys.path` first.
- `pytest tests/` could not run **at all**: `tests/test_security_fix.py` imports helpers removed by `08fd99ba`, and one collection error aborts the entire run.
- The plan-revision cap could never fire — the wired lambda read `revisions_count`, a key nothing writes, while the real router reading `plan_revision_count` sat unused (#2062).
- All three review routers raise at their ceiling, but nothing caught them, so hitting a cap crashed instead of degrading.

## What this includes

| | |
|---|---|
| Ship-blocker fixes | #1943 #2063 #2072 #2051 |
| CI | new `tests.yml` — `imports` (3.12/3.13/3.14), `collect`, `unit` |
| Issue-backed community fixes | 17 PRs, each tested individually before stacking |
| `@Bartok9` consolidation | 67 evaluated → 46 kept → 5 per-area commits |

**Resolves 97 of 129 open PRs** (68 merged, 29 superseded) and **17 open issues**.

## Verification

```
real Python 3.12 : 387 passed, 2 skipped, 0 failed
real Python 3.14 : 387 passed, 2 skipped, 0 failed
pip install -e ".[test]" on clean 3.12: OK
121/121 submodules import · ASGI app boots · multi_agents graph builds
```

Live A/B research run, default Tavily retriever, same query:

| | `main` | this branch |
|---|---|---|
| cost | $0.1532 | $0.1535 |
| sources | 14 | 13 |
| context | 40,966 chars | 40,865 chars |
| elapsed | 76s | 80s |

No new content filter fired on real traffic. The report chat RAG path — dead code (`and False`) until now — was exercised end to end: vector store and retriever build in ~1s and answer correctly.

Baseline before this branch: **46 tests, and `pytest tests/` could not run.**

## Behaviour changes worth knowing

- **`ALLOW_PRIVATE_URLS`** — the new SSRF guard resolves each scraped URL and rejects private, loopback, link-local and metadata addresses. **Anyone researching internal hosts must opt back in.** Documented in `.env.example`.
- **Report chat now builds a vector store**, so chat sessions make embedding calls they previously did not. Wrapped in a no-RAG fallback.
- **`FIRECRAWL_CONCURRENCY` defaults to 2**, matching the free tier. Paid-plan users should raise it.
- `asyncio_mode` moves `strict` → `auto`; three existing async tests carry no marker and error under `strict`.

## Deliberately not included

- **#1982** — fails its own four tests and breaks two existing ones against this base. Needs author rework.
- **#1756** — only the FireCrawl and config hunks; the branch also carries unrelated docs churn.
- `--forked` is required in CI: the suite still leaks global state across modules (14 failures without it, unchanged from `main`). Fixing that leakage is the prerequisite for dropping the flag.

## A note on the `@Bartok9` tests

Several stub `gpt_researcher` into `sys.modules` at import time and never restore it, breaking collection for every later module. Two did `sys.modules.setdefault("bs4", ...)` followed by an unconditional `sys.modules["bs4"].BeautifulSoup = MagicMock` — `setdefault` returns the **real** module when bs4 is already imported, so the real `BeautifulSoup` became a `MagicMock` process-wide. `--forked` does not help, since collection happens in the parent. Six files were repaired here; future submissions should be screened for this.


Bartok9:
Sorry to ping a merged PR — @FlorianFranzen flagged on #1989 that `39215c5b` credits Claude Opus 5 as Co-Authored-By but not the community authors whose PRs were stacked (including #1989).

If it is still easy to add trailers on a follow-up, please include:

```
Co-Authored-By: Bartok9 <259807879+Bartok9@users.noreply.github.com>
```

Thank you for landing the batch. Full note is on #1989.

PR #2078, opened 2026-08-23
fix(scraper): keep extracted content when web_base_loader enrichment fetch fails

### LINKED ISSUE #2077: WebBaseLoaderScraper discards extracted content when image/title enrichment fetch fails
`WebBaseLoaderScraper.scrape()` extracts page content via `WebBaseLoader.load()`, then makes a second, independent request purely for image/title enrichment (`self.session.get(self.link)` at line 31). Both the primary extraction and the enrichment fetch sit inside the same `try` block, and the blanket `except Exception` at the end returns `"", [], ""`, discarding the content that was already extracted successfully.

Any failure in the enrichment step wipes real content: a transient `ConnectionError`/`SSLError` on the second request, a rate limit or WAF block hit only on that second round trip, or a parse failure inside `get_relevant_images`/`extract_title` on a malformed page.

Repro (isolates `web_base_loader.py` with `WebBaseLoader.load()` stubbed to return real content, and `self.session.get` raising a `ConnectionError`):

```
content, images, title = scraper.scrape()
# content == "" even though WebBaseLoader.load() returned real page content
```

This is the same bug shape already fixed in `beautiful_soup.py` (#1684) and that other scraper backends (`firecrawl.py`, `tavily_extract.py`) have recently had fixed too: the enrichment fetch should be allowed to fail independently of the already-successful content extraction.

PR incoming: split the enrichment fetch into its own inner `try`/`except` so a failed image/title lookup degrades to empty `image_urls`/`title` instead of discarding `content`.

AmirF194:
Fixes #2077

`WebBaseLoaderScraper.scrape()` extracts content with `WebBaseLoader.load()`, then makes a second request purely for image/title enrichment (`self.session.get`). Both sat in the same `try` block, so a failure in the enrichment step (a transient network error, a rate limit hit only on the second round trip, a parse failure on a malformed page) fell into the blanket `except` and discarded content that had already been extracted successfully.

The fix wraps the enrichment fetch and parse in its own `try`/`except`, so a failure there only leaves `image_urls`/`title` empty instead of wiping `content`. Same split `beautiful_soup.py` already uses for the same class of bug.

Verification:
- New `tests/test_web_base_loader_enrichment_guard.py`: fails on `main` (content wiped to `""`), passes on this branch (content preserved, `image_urls`/`title` empty), for both a raised exception on the fetch and one raised during parsing.
- Full offline suite (`pytest tests/ --forked --timeout=60`, deselecting the three tests that need live credentials/network per the repo's own CI config): 391 passed, 1 skipped, matching `main`.
- `imports` and `collect` CI jobs reproduced locally, both clean.

### DIFF
--- a/gpt_researcher/scraper/web_base_loader/web_base_loader.py
+++ b/gpt_researcher/scraper/web_base_loader/web_base_loader.py
@@ -34,12 +34,16 @@ class WebBaseLoaderScraper:
                     continue
                 content += str(page)
 
-            response = self.session.get(self.link)
-            soup = BeautifulSoup(response.content, 'html.parser')
-            image_urls = get_relevant_images(soup, self.link)
-            
-            # Extract the title using the utility function
-            title = extract_title(soup)
+            image_urls, title = [], ""
+            try:
+                response = self.session.get(self.link)
+                soup = BeautifulSoup(response.content, 'html.parser')
+                image_urls = get_relevant_images(soup, self.link)
+
+                # Extract the title using the utility function
+                title = extract_title(soup)
+            except Exception as e:
+                print("Error extracting images/title! : " + str(e))
 
             return content, image_urls, title
 
--- a/tests/test_web_base_loader_enrichment_guard.py
+++ b/tests/test_web_base_loader_enrichment_guard.py
@@ -1,0 +1,98 @@
+"""WebBaseLoaderScraper keeps already-extracted content when image/title enrichment fails."""
+
+from __future__ import annotations
+
+import importlib.util
+import sys
+import types
+import unittest
+from pathlib import Path
+from unittest.mock import MagicMock
+
+
+ROOT = Path(__file__).resolve().parents[1]
+MODULE_PATH = ROOT / "gpt_researcher" / "scraper" / "web_base_loader" / "web_base_loader.py"
+
+
+def _load(get_relevant_images=lambda soup, link: []):
+    bs4 = types.ModuleType("bs4")
+
+    class BeautifulSoup:
+        def __init__(self, *a, **k):
+            pass
+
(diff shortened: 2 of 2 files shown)


assafelovic:
Reviewed and merging — good catch, and a clean fix.

Verified against current `main`: applies cleanly, full suite passes (393 tests, 0 failures, your two included).

The bug is a real one and nicely characterised: `WebBaseLoaderScraper` had already successfully loaded the document, then threw away that content because a **secondary** enrichment fetch for images and title failed. Wrapping only the enrichment and defaulting to `([], "")` keeps the useful work and degrades the decoration, which is the right shape.

One note on the test, not blocking. `_load()` assigns `sys.modules["bs4"]` and `sys.modules["gpt_researcher"]` unconditionally and never restores them. It happens to be harmless today — those assignments run at *test* time, after collection has already bound the real modules elsewhere — and CI runs the suite under `--forked`, so each test is isolated.

Worth knowing it's a landmine in this repo specifically: #2074 had to repair six test files doing exactly this. Two of them did `sys.modules.setdefault("bs4", ...)` followed by an unconditional `sys.modules["bs4"].BeautifulSoup = MagicMock` — and since `setdefault` returns the *real* module when bs4 is already imported, that replaced the real `BeautifulSoup` with a `MagicMock` process-wide. Others left `gpt_researcher` stubs in `sys.modules` permanently and broke collection for every module after them.

Your `setdefault` on `requests` is the safe form. If you ever extend this file, snapshotting `sys.modules` and restoring at the end is cheap insurance.

AmirF194:
Thanks for the fast review and the detail on the sys.modules landmine, good to know for next time. I kept setdefault for that exact reason but had not seen the MagicMock replacement failure mode from #2074, will snapshot and restore if I touch this file again.

PR #2092, opened 2026-08-30
fix: define stream output for multi-agent execution

alaningtrump:
The multi-agent WebSocket helper passes stream_output to run_multi_agent_task, but the callback was not defined in server_utils.

Import the existing actions callback so the execution path can start normally. No API behavior changes outside this error path.


alaningtrump:
@assafelovic Could you please take a look when you have a chance? Thanks!

PR #2111, opened 2026-09-08
fix: handle zero generated queries in deep research

### LINKED ISSUE #2110: Deep research zero-query path raises NameError before fallback state is initialized
## Summary

`DeepResearchSkill.deep_research()` has a zero-query guard immediately after `generate_search_queries()`, but the guard returns variables that are initialized only **after** the guard.

On current `main`, the control flow is effectively:

```python
serp_queries = await self.generate_search_queries(query, num_queries=breadth)

if not serp_queries:
    return {
        "learnings": all_learnings,
        "visited_urls": all_visited_urls,
        "citations": all_citations,
        "context": all_context,
        "sources": all_sources,
    }

all_learnings = learnings.copy()
all_citations = citations.copy()
all_visited_urls = visited_urls.copy()
all_context = []
all_sources = []
```

Therefore, if query generation legitimately returns `[]`, the intended graceful-stop branch instead raises `NameError` because `all_learnings` (and the other `all_*` values) do not yet exist.

## Why this matters

The function explicitly tries to make the zero-query case a safe termination path:

```python
logger.warning("Deep research generated zero search queries; stopping descent.")
```

That condition can occur if the model returns malformed/empty output and `parse_search_queries_response()` yields no valid queries. In that situation, deep research should preserve the accumulated state and stop cleanly rather than failing in the fallback itself.

## Expected behavior

If `generate_search_queries()` returns no queries, `deep_research()` should return the current accumulated state:

- existing `learnings`;
- existing `citations`;
- existing `visited_urls`;
- empty context/sources for the current level.

No exception should be raised.

## Actual behavior

The zero-query branch references these names before assignment:

- `all_learnings`
- `all_visited_urls`
- `all_citations`
- `all_context`
- `all_sources`

The first reference raises `NameError`.

## Suggested fix

Initialize the accumulated-state variables before generating/checking the query list:

```python
all_learnings = learnings.copy()
all_citations = citations.copy()
all_visited_urls = visited_urls.copy()
all_context = []
all_sources = []

serp_queries = await self.generate_search_queries(query, num_queries=breadth)

if not serp_queries:
    logger.warning("Deep research generated zero search queries; stopping descent.")
    return {
        "learnings": all_learnings,
        "visited_urls": all_visited_urls,
        "citations": all_citations,
        "context": all_context,
        "sources": all_sources,
    }
```

## Regression coverage

A focused async unit test can stub `generate_search_queries()` to return `[]` and verify that:

1. no exception is raised;
2. pre-existing learnings/citations/visited URLs are preserved;
3. current-level `context` and `sources` are empty;
4. no nested researcher or retrieval work is started.

I searched the current issue set for this specific zero-query / pre-initialization failure and did not find an existing report for the same root cause.

<sub>AI-assisted review disclosure: I used an AI coding assistant to help inspect the control flow and draft this report; the reported behavior is based on the current source structure and is intentionally scoped to the pre-initialization bug above.</sub>

anupamking01:
## Summary

Fixes the zero-query fallback in `DeepResearchSkill.deep_research()` so it can terminate cleanly when query generation returns no valid queries.

Previously, the guard returned `all_learnings`, `all_citations`, `all_visited_urls`, `all_context`, and `all_sources` before those variables were initialized, causing a `NameError` instead of preserving the accumulated research state.

## Changes

- initialize accumulated deep-research state before generating/checking search queries;
- preserve existing learnings, citations, and visited URLs when query generation yields `[]`;
- add an async regression test covering the zero-query path and confirming no context/sources are fabricated.

Fixes #2110.

<sub>AI-assisted contribution disclosure: an AI coding assistant was used to help inspect the control flow and draft the patch/test; the change is intentionally scoped to the reported bug.</sub>


anupamking01:
Quick review note: this PR is still mergeable and intentionally scoped to the zero-query fallback only. The regression test is fully offline (query generation is stubbed to return `[]`) and checks that existing learnings/citations/visited URLs are preserved without starting nested retrieval. The Actions runs are currently showing `action_required`, which appears to need upstream approval before fork workflows can execute. Happy to make any changes the maintainers prefer.

